PR #235, opened 2014-12-10
Additional apt-get update to fix key errors found

hkumarmk:
1. The initial failures are happening because of package installation failures
because of apt key issues - apt-get install throwing warnings "packages cannot
be authenticated". A manual apt-get update fixed that problem (I tried it in two
different acceptance test environments).

So added a fix for that - if apt-get install fail, repeat apt-get update.

2. Repeating puppet apply was not working because of a typo (missing escape on
collecting return value. Fixed it by escaping it.

### DIFF
--- a/build_scripts/deploy.sh
+++ b/build_scripts/deploy.sh
@@ -37,11 +37,17 @@ then
        dpkg -i internal.deb
 fi
 apt-get update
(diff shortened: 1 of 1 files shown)


bodepd:
@hkumarmk this either needs to be rebased or closed

bodepd:
I'm going to close this b/c of lack of activity. please rebase and reopen if you still think it's necessary.

PR #270, opened 2014-12-18
Make sure we have rjil-cicd installed

sorenh:


sorenh:
This happens due to dependency resolution anyway. No longer needed.

PR #289, opened 2015-01-05
Add ssh keys for himanshu from QA Team

pandeyop:
This is just going to be a test build which QA Team members would use. Once the environment from this build is up the older environment being used by QA Team would be deleted.


pandeyop:
please test this

sorenh:
It seems that sometimes Jenkins won't start these builds if you do "please test this" before the Travis job is done.

Try again and see if works this time.

sorenh:
Weird. It started just now. Maybe my comment caused Jenkins to see your old comment.

PR #293, opened 2015-01-06
Stop using Contrail PPA

sorenh:
Contrail PPA was updated to 2.0. We're not quite ready to make that switch yet and our stuff stopped working, because we were mixing 1.10 with 2.0. Stop using the PPA and use the rustedhalo stuff instead. I've filled in the gaps in rustedhalo to make this work again.


sorenh:
please test this

bodepd:
please test this

sorenh:
please test this

sorenh:
please test this

PR #313, opened 2015-01-15
Configure vagrant VM resources using layouts and maps

hkumarmk:
This patch makes vagrant to see the resources mentioned in the layout yaml file
(e.g. environments/full.yaml) and then map to map yaml file (e.g
environments/vagrant-vbox.yaml) for vm resources. So vagrant will be following
somewhat similar method for resource configuration in other environments.

This is also useful to build a hybrid vagrant environment which contain vms and
containers for appropriate machines.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,18 +2,19 @@ require 'yaml'
 
 Vagrant.configure("2") do |config|
 
+  default_flavor = 'small'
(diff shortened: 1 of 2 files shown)


hkumarmk:
This patch only affect vagrant environment. So no need for integration testing.

@bodepd Could you please take a look at it.

PR #315, opened 2015-01-15
Skip Tempest Tests not relevant to current environment

pandeyop:
We can skip tempest tests which aren't currently relevant for our environment. This would limit failures to reflect only those components which we care about "at the moment".
The tests have been skipped through the below negative lookahead regex: 
./run_tempest.sh -N 'tempest\.*(?!(.*thirdparty)|(.*baremetal)|(.*data_processing)|(.*messaging)|(.*orchestration)|(.*vpnaas)|(.*test_metering)|(.*cli)|(.*test_load_balancer)|(.*fwaas))'

This can be modified to include/exclude more tests as and when more components are enabled in our environment.

### DIFF
--- a/build_scripts/test.sh
+++ b/build_scripts/test.sh
@@ -4,4 +4,4 @@
 
 python -m jiocloud.apply_resources ssh_config --project_tag=${project_tag} ${mappings_arg} environment/${layout:-full}.yaml > ssh_config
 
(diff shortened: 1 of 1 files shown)


hkumarmk:
please test this

hkumarmk:
I see tempest run is working there. Merging this.

bodepd:
@hkumarmk I looked at the output and it looks like something failed...

hkumarmk:
@bodepd as discussed, this is happening on other builds also for at least last two days

PR #317, opened 2015-01-15
Install jiocloud-ssl-certificate package on all machines

hkumarmk:
This patch install jiocloud-ssl-certificate package on all machines as part of
trust_selfsigned_cert.

In order to connect to the api endpoints (without insecure option), all machines
need to trust selfsigned certificate.

This should fix puppet failures on some nodes (like etcd) on vagrant
environment.

### DIFF
--- a/manifests/trust_selfsigned_cert.pp
+++ b/manifests/trust_selfsigned_cert.pp
@@ -3,8 +3,14 @@
 #
 
 class rjil::trust_selfsigned_cert(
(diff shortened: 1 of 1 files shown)


bodepd:
I'm not 100% convinced by this. The ssl certs need to be installed to deploy the right secrets, so it seems like something that needs to be done before Puppet runs. Does it make sense to have it in Puppet?

bodepd:
please test this

PR #320, opened 2015-01-15
ENV['env'] will be true, hence the second condition is not going.

saneax:
expliciting set map value if env['map'] is false.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -10,7 +10,7 @@ Vagrant.configure("2") do |config|
   # cloud file that is used
   environment = ENV['env'] || 'vagrant-vbox'
   layout = ENV['layout'] || 'full'
-  map = ENV['map'] || ENV['env']
+  map = ENV['map'] || 'vagrant-vbox'
 
   config.vm.provider :virtualbox do |vb, override|
     vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]


liferich:
What will this change do? It is supposed to be either map or env variable.
env environment anyway have vagrant-box or anything like that. Why do we need to hardcode it as vagrant-vbox?

In future we are going to have hybrid vagrant environment in which case vbox wont make sense in the env.

Vagrantfile
```diff
@@ -10,7 +10,7 @@ Vagrant.configure("2") do |config|
   # cloud file that is used
   environment = ENV['env'] || 'vagrant-vbox'
   layout = ENV['layout'] || 'full'
-  map = ENV['map'] || ENV['env']
+  map = ENV['map'] || 'vagrant-vbox'
```
bodepd: map should really be cloud_provider which has not been added to the vagrant stuff. this should perhaps be environment, but definitely not hard-coded

hkumarmk:
sure itll be changed to cloud_provider.

bodepd:
after some discussion this week, I think it's clear that we are going to deprecate vagrant support ASAP. because of this, I really only want to fix things that are blockers

saneax:
just for the sake of omitting -
map=vagrant-vbox for now, please merge this. I am having to run Vagrant via 
sanjayu@cmpt:~/workspace/puppet-rjil$ export map=vagrant-vbox
sanjayu@cmpt:~/workspace/puppet-rjil$ vagrant status

hkumarmk:
@saneax why do you need to set map? you can do it by just setting up env=vagrant-vbox or even without setting anything (env is defaults to vagrant-vbox and if you dont set map environment variable, itll just use vagrant-vbox.

hkumarmk:
im sorry you have to set environment variable env as it is used by puppet also.

bodepd:
for now I'm going to close this as it won't be merged in it's current form.

PR #331, opened 2015-01-17
removing dependency on httpd for cinde,glance,neutron,nova

saneax:
On Vagrant we were seeng oc1 (keystone) failure on dependency for (say) cinder,
which was again failing on dependency for httpd, and httpd was failing on dependency for keystone

### DIFF
--- a/manifests/cinder.pp
+++ b/manifests/cinder.pp
@@ -86,8 +86,6 @@ class rjil::cinder (
 
   include rjil::apache
 
-  Service['cinder-api'] -> Service['httpd']
-
   ##
   # Cinder module dont have bind port parameter, so adding here for now.
   ##
--- a/manifests/glance.pp
+++ b/manifests/glance.pp
@@ -38,9 +38,6 @@ class rjil::glance (
 
   include rjil::apache
 
-  Service['glance-api'] -> Service['httpd']
-  Service['glance-registry'] -> Service['httpd']
-
   ## Configure apache reverse proxy
   apache::vhost { 'glance-api':
     servername      => $server_name,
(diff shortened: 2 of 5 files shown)


saneax:
please test it

saneax:
please test this

bodepd:
@saneax 

two problems with this patch that I can see: 

== missing dependency is not specified:

all keystone objects need to come only after Service[http] is started

== commit messages should have enough context for anyone reviewing code to understand:

* what was the problem (ie: what problem was occurring)
* what does the fix do

if I had not been sitting with you debugging, I could not really understand the intention of this patch without really digging into the code.

@hkumarmk for context when you review:

basically, there is a missing dependency between:

```
Service['http'] -> Keystone_obj<||>
```

in fact, there are deps to ensure that the keystone stuff has to come after keystone is installed b/c keystone_obj -> service-api -> http (and the keystone calls cannot work without http running)

for the case of acceptance, it only happened to work b/c the --tags package run from userdata happened to install apache.

bodepd:
from a functional perspective, the patch does appear to result in a working cluster

saneax:
please test this

saneax:
The test has failed due to timeout, not running a new one.

saneax:
please test this

hkumarmk:
The reason why I put the order is  that, restarting apache with vhost configuration before (re)configuring api services will be failed, as  vhost will have same port configuration as api services (we are actually moving the default api ports to apache vhosts), will need to analyze this more to comment more on this.

anshprat:
closing this as not required.

PR #338, opened 2015-01-20
WIP: Run travis jobs in containers

theanalyst:
This a series of commits to enable running of Travis jobs in their container based infrastructure. Currently I'm just putting it up here for review, there may be better ways to do the workarounds I've done to get this up ( improvements welcome)

- `bodeco_module_helper` reference is updated to use the gem which ignores the Lint tasks in vendor dir, (this may need to change as the commit doesn't refer master). However this doesn't completely solve the issue as there is a bug in puppet lint https://github.com/rodjek/puppet-lint/issues/331 , which ignores the ignore paths (no pun intended :)), so the Rakefile kind of works around this with the fix mentioned in the issue

- puppet-syntax also fails in vendor directories, and this is also worked around by explicitly adding PuppetSyntax exclude paths in the rake file.

- *.pp files in spec/classes dir are renamed to *.rb as travis lint jobs also fail here 

- puppet-lint.rc is added to explicitly pass on `autoloader_layout-check`

PS: as an aside, travis runs complete about ~10 mins faster due to the faster run in containers & bundler caching


.puppet-lint.rc
```diff
@@ -0,0 +1 @@
+--no-autoloader_layout-check
```
bodepd: I think we should fix this issue instead of disabling these checks b/c they are valuable

theanalyst: Yeah, figured how to remove it, config.relative was failing due to another issue (
https://github.com/rodjek/puppet-lint/commit/e6e0d4fd1ab0fb6415e5ade0958210ad944b0087
fixed it

Either we could track master in Gemfile or alternatively adding one more line to Rakefile 
ie `PuppetLint.configuration.relative = true` should fix this.

theanalyst:
@bodepd removed the lintrc file, fixed Rakefile to do the relative autoloader option

PR #347, opened 2015-01-20
Removed glance image-list from glance check

ajayaa:


bodepd:
I agree this patch needs to happen, but as of now, it should probably wait until a refactor that I am still working on in my local repo. I just want to make sure that we don't actually lose the glance-image check (it needs to move to the validation checks so that it still at least triggers a service warning if it is not functional)

bodepd:
this patch should be replaced by #360 which achieves the same thing as a part of a service check refactor.

bodepd:
this should not be required anymore b/c #360  has been merged, feel free to re-open if you think it still makes sense.

PR #348, opened 2015-01-21
Removed service blocker to avoid unneccessary wait

hkumarmk:
* Removed service blocker before keystone objects
* Added dependancy on http for keystone

This should result a working openstack cluster in vagrant. It may need further
optimization, but this should work for now.

### DIFF
--- a/manifests/keystone.pp
+++ b/manifests/keystone.pp
@@ -19,6 +19,8 @@ class rjil::keystone(
   $disable_db_sync        = false,
 ) {
 
+  Service['keystone'] -> Service['httpd']
+
   if $public_address == '0.0.0.0' {
     $address = '127.0.0.1'
   } else {
--- a/site.pp
+++ b/site.pp
@@ -109,11 +109,6 @@ node /^ocdb\d+/ {
   include rjil::openstack_zeromq
   include openstack_extras::keystone_endpoints
   include rjil::keystone::test_user
-  # ensure that we don't create keystone objects until
-  # the service is operational
-  ensure_resource('rjil::service_blocker', 'keystone-admin', {})
-  Rjil::Service_blocker['keystone-admin'] -> Class['openstack_extras::keystone_endpoints']
-  Rjil::Service_blocker['keystone-admin'] -> Class['rjil::keystone::test_user']
 }
 
 #


bodepd:
what ensures that the keystone objects don't get configured until the keystone api is up?

hkumarmk:
@bodepd I am still working on this.

hkumarmk:
Basically the service blocker there in site.pp doesnt solve the purpose without complex deps as the other required resources (mentioned below) required to create keystone objects are getting created after the service blocker (whether it is on same machine or different machines) and thus service blocker is timing out and failing first execution (which should be okay as the keystone objects creation will just fail because of the endpoints are not ready). But during second or third puppet run, it just work and giving a working cluster.

The current (required) deps within ocdb1:

* keystone -> apache -> <keystone-admin.service.consul up > -> dns config for identity.jiocloud.com -> keystone object cretation. 
* stmon.service.consul -> mon_config -> cinder-volume
* keystone service for cinder and mon_config -> cinder-api -> apache
* stmon.service.consul and keystone service for glance -> mon_config -> glance -> apache
* memcache -> runtime_fail(fail_before_nova) -> nova -> apache
* keystone service for nova -> nova api

Here apache vhost  should have dependant on individual services, because currently apache vhost is getting configured regardless of failures of any dependency for the openstack services which is causing failure of apache service start because of port conflict during first puppet run - fixing this would avoid delay on starting apache and keystone.

manifests/glance.pp
```diff
@@ -38,8 +38,8 @@
 
   include rjil::apache
 
-  Service['glance-api'] -> Service['httpd']
-  Service['glance-registry'] -> Service['httpd']
+  Service['glance-api'] -> Apache::Vhost<|title == 'glance-api'|>
+  Service['glance-registry'] -> Apache::Vhost<|title == 'glance-registry'|>
```
bodepd: what's the difference between this and just not having the dependencies?

hkumarmk:
If we dont have dependencies, vhost will be configured even though cinder/glance/nova fail to configure because of other dep failures, and when apache service try to be refreshed, apache will fail to start because of port conflicts (api services would be started in their default ports during package installation) - this will also affect keystone even though keystone come up without issues. 

Keeping the vhost dependent on the api services, the vhosts will not be configured until api services are up, and any failures on api services because of external dependencies during startup, will leave apache healthy with only successful api services like keystone. 

This will allow keystone to come early - in first iteration itself, otherwise it may be completely configured second or third iteration.

bodepd:
@hkumarmk thanks for that clarification, makes sense to me

saneax:
Please merge this, I am hitting this issue in vagrant deployment

bodepd:
just for clarification, this is definitely not prioritized at the moment, @hkumarmk and I are pretty busy working on getting services set up with HA/horizontal scaling this week.

PR #361, opened 2015-01-23
Added a method to get ip of node

sajuptpm:
build_scripts/deploy.sh uses the statement
"python -m jiocloud.utils get_ip_of_node" in multiple
places, So added a function in common.sh to pack it.


bodepd:
my instinct is to reject and close this. Having a method in bash to wrap a single call to something that has already been encapsulated as a python command line invocation seems like too much extra code for no value. Feel free to disagree if you feel passionate about this patch. /cc @sorenh @hkumarmk in case you guys want to disagree.

PR #365, opened 2015-01-26
Add a hwprofile layer to hiera

sorenh:
In the new production environment, there are more than one server type.
This means there are different names for interfaces, so they need
different values for e.g. public_interface. Although there's some
correlation between server type and role, this is not reliable.
    
I'm hoping that the productname will be a sufficiently strong identifier
for these kinds of settings. If not, we'll have to revisit in the
future.

### DIFF
--- a/hiera/hiera.yaml
+++ b/hiera/hiera.yaml
@@ -10,6 +10,7 @@
   - services
   - "secrets/%{env}"
   - "role/%{jiocloud_role}"
+  - "hw/%{cloud_provider}/%{productname}"
   - "cloud_provider/%{cloud_provider}"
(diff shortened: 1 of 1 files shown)


bodepd:
uggh that's ugly, but I totally get it...

bodepd:
is there a custom fact to set productname?

sorenh:
Uh, no? It's just there.

sorenh:
@bodepd Are you suggesting you don't have a productname fact?

hkumarmk:
I see the fact there on servers.

hkumarmk:
merging this.

PR #379, opened 2015-01-29
Use "combo" as default key name

hkumarmk:
combo key include public keys of both Soren, Dan and Harish and is added there
on all production, hp cloud.


bodepd:
please test this

PR #380, opened 2015-01-29
Workaround to add swap partition on baremetal nodes

hkumarmk:
Cloudinit is creating the swap partition, its not added to the fstab and not
enabled. This patch will add that steps to userdata so this will be done during
boot.


sorenh:
I would really much rather see this as a patch against cloud-init.

bodepd:
please test this

bodepd:
this is ready to be merged

bodepd:
please test this

PR #381, opened 2015-01-29
Do not use pip to install tempest-lib package

pandeyop:
Stop using pip to install tempest-lib package. 
Switch to python-tempest-lib package from the repo.
Note: We need tempest-lib:0.1.0 to be able to run tempest tests.


sorenh:
please test this

PR #382, opened 2015-01-29
Display that we are retrying rather than timing out. And that it is expected

rickysarraf:
Display that we are retrying rather than timing out. And that it is expected


bodepd:
this needs to be rebased.

PR #394, opened 2015-02-06
Rebase:  Display that we are retrying rather than timing out. 

rickysarraf:
Rebase:  Display that we are retrying rather than timing out. And that it is exptected...


bodepd:
needs to be rebased.

bodepd:
I like the idea of this patch, but there are already so many log lines on the screen during this process, that I don't want to add any more lines. Can someone else jump in?

PR #403, opened 2015-02-09
Retry on wget

### LINKED ISSUE #402: proxy node sometimes not reachable on gatetests
Sometimes proxy node is not reachable from the other nodes temporarily on gate tests, but retry proxy node after some time is working. This is causing gate failures

hkumarmk:
Noticed good number of gate test failures because of temporary connection issues
to proxy node. So adding retry on wgets in userdata.

Also this patch add a retry function which can be reused.

This should fix #402


bodepd:
please test this

bodepd:
wget provides this capability from the command line. closing.

PR #404, opened 2015-02-09
Sleep long enough for cloud-init to have a chance to give up

sorenh:
If DHCP fails, that'll take 2 minutes before the networking upstart
job gives up. Then another 2 minutes for cloud-init to give up reaching
the MD service. Add a buffer of an extra 30 seconds. That makes 270
seconds.


sorenh:
I took a look at the AT runs and realised that with only 100 seconds of sleep, it's pointless since cloud-init doesn't give up until after 120 seconds.

sorenh:
Ta very much.

PR #412, opened 2015-02-11
WIP: Cleanup Vagrant and make it use userdata.txt

saneax:
1. smoothing out Vagrantfile to run "puppet apply --tags packages first"
2. Adding a default sources.list, this does the initial package pull faster. (WIP)
3. adding provision to move to a configurable mirror

### DIFF
--- a/build_scripts/make_userdata.sh
+++ b/build_scripts/make_userdata.sh
@@ -9,6 +9,12 @@ export LANG=en_US.UTF-8
 export LANGUAGE=en_US.UTF-8
 export layout="${layout}"
 release="\$(lsb_release -cs)"
+consul_discovery_token=${consul_discovery_token:-"UNSET"}
+BUILD_NUMBER=${BUILD_NUMBER:-"UNSET"}
+env=${env:-"UNSET"}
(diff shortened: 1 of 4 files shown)


saneax:
WIP, so please do not merge

saneax:
please test this

saneax:
please test this

bodepd:
this patch conflates way too many changes. I would prefer if you can isolate the changes that only effect vagrant from those that potentially effect all builds (ie: things that touch userdata) and submit them separately.

build_scripts/make_userdata.sh
```diff
@@ -38,7 +54,13 @@ then
        dpkg -i internal.deb
 fi
 apt-get update
-apt-get install -y puppet software-properties-common puppet-jiocloud jiocloud-ssl-certificate
+
+if [ "${cloud_provider}" == "hp"]; then
```
bodepd: why would we only install certain packages in hp cloud and not for example when we move to our own cloud for testing? I would prefer if we limit the amount of per cloud differences as much as possible.

Issue #407, opened 2015-02-09
puppet from userdata and cron should never run at the same time


bodepd:
Currently, since the userdata run of puppet retries, it is possible to wind up in a situation where both calls (cron and userdata ) run at the same time.

You should not run Puppet at the same time b/c it can lead to locking issues with packages (as well as other thing b/c it is not a supported use case). I think we can set lock files to ensure only one runs at a time from puppet apply (as in, I think these are supported command line options).

sorenh:
We already have a locking mechanism for this (maybe-upgrade runs under run-one). Perhaps we can reuse it?

saneax:
The userdata part of while loop for puppet apply, should have 'maximum nos of run', which we can configure via a variable, rather than an endless loop. Most often if it fails a certain no of times, it usually will fail throughout and failing the jenkins job.

stefanandres:
A practical workaround would be:
- wrap the puppet runs inside a wrapper which does the locking
 - call the wrapper via cron and in bootstrapping

PR #413, opened 2015-02-11
WIP: mods to make userdata.txt more generic

saneax:
1. Documenting variables important for this run
2. if cloud_provider == hp, then install all the puppet specific stuff from git, including puppet modules, else (assuming) vagrant and we have puppet mounted already. We are also assuming these modules are latest ie as per (Puppetfle)


saneax:
I will run gate tests after the vagrant tests pass on my local vagrant setup.

build_scripts/make_userdata.sh
```diff
@@ -9,6 +19,13 @@ export LANG=en_US.UTF-8
 export LANGUAGE=en_US.UTF-8
 export layout="${layout}"
 release="\$(lsb_release -cs)"
+
+#populating safe values (UNSET) for debugging
```
bodepd: any reason this can't just use common.sh?

build_scripts/make_userdata.sh
```diff
@@ -27,9 +44,9 @@ fi
 wget -O puppet.deb -t 5 -T 30 http://apt.puppetlabs.com/puppetlabs-release-\${release}.deb
 if [ "${env}" == "at" ]
 then
-	jiocloud_repo_deb_url=http://jiocloud.rustedhalo.com/ubuntu/jiocloud-apt-\${release}-testing.deb
+       jiocloud_repo_deb_url=http://jiocloud.rustedhalo.com/ubuntu/jiocloud-apt-\${release}-testing.deb
```
bodepd: what is the reason for this whitespace change?

build_scripts/make_userdata.sh
```diff
@@ -27,9 +44,9 @@ fi
...
 else
-	jiocloud_repo_deb_url=http://jiocloud.rustedhalo.com/ubuntu/jiocloud-apt-\${release}.deb
+       jiocloud_repo_deb_url=http://jiocloud.rustedhalo.com/ubuntu/jiocloud-apt-\${release}.deb
```
bodepd: same question here

build_scripts/make_userdata.sh
```diff
@@ -38,48 +55,57 @@ then
...
-  if [ -n "${pull_request_id}" ]; then
-    pushd /tmp/rjil
-    git fetch origin pull/${pull_request_id}/head:test_${pull_request_id}
-    git config user.email "testuser@localhost.com"
-    git config user.name "Test User"
-    git merge -m 'Merging Pull Request' test_${pull_request_id}
-    popd
+
+#vagrant mounts puppet manifest paths, and cloud installs puppet-jiocloud package.
+if [ "${cloud_provider}" == "hp"]; then
```
bodepd: what does this have to do with hp cloud? Would we not want to run this same code if we were using rackspace cloud for tests or our own internal cloud?

build_scripts/make_userdata.sh
saneax: I have removed this now.

build_scripts/make_userdata.sh
saneax: so, what I see in the already existing source, few places tabs are expanded to whitespaces and few places its \t. What is accepted norm? my vi turns tabs to spaces by default. What I see as well is we do not have a standard on tabs/spaces and how many. 
lines 13 has 2 space for a tab
lines 19,20 and 24,25,30,32has a tab (which is equal to 8 spaces)
line 38 is 8 spaces :) and it goes on.

build_scripts/make_userdata.sh
```diff
@@ -72,29 +92,34 @@ if [ -n "${puppet_modules_source_repo}" ]; then
...
-    ret_code_jio=\$?
-    if [[ \$ret_code_jio = 1 || \$ret_code_jio = 4 || \$ret_code_jio = 6 || \$ret_code_package = 1 || \$ret_code_package = 4 || \$ret_code_package = 6 ]]
-    then
-        echo "Puppet failed. Will retry in 5 seconds"
-        sleep 5
-    else
-        break
-    fi
+  # first install all packages to make the build as fast as possible
+  run-one puppet apply --detailed-exitcodes \`puppet config print default_manifest\` --tags package
```
saneax: trying to address #407 here.

build_scripts/make_userdata.sh
bodepd: we should be expanding to whitespace.

saneax:
please test this

build_scripts/make_userdata.sh
saneax: we will correct all  the indents to whitespace, however it would look like
a new file. should I do it with this change?

On Thu, Feb 12, 2015 at 12:51 PM, Dan Bode <notifications@github.com> wrote:

> In build_scripts/make_userdata.sh
> <https://github.com/JioCloud/puppet-rjil/pull/413#discussion_r24562741>:
>
> > @@ -27,9 +44,9 @@ fi
> >  wget -O puppet.deb -t 5 -T 30 http://apt.puppetlabs.com/puppetlabs-release-\${release}.deb
> >  if [ "${env}" == "at" ]
> >  then
> > -	jiocloud_repo_deb_url=http://jiocloud.rustedhalo.com/ubuntu/jiocloud-apt-\${release}-testing.deb
> > +       jiocloud_repo_deb_url=http://jiocloud.rustedhalo.com/ubuntu/jiocloud-apt-\${release}-testing.deb
>
> we should be expanding to whitespace.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/JioCloud/puppet-rjil/pull/413/files#r24562741>.
>



-- 
Sanjay Upadhyay
http://saneax.blogspot.com

build_scripts/make_userdata.sh
```diff
@@ -38,13 +48,23 @@ then
...
+
+#Below Packages should be common to all env's (vagrant/hp/etc)
+#we are adding run-one so that we can run puppet repeatedly with this, solving issue #407
+apt-get install -y puppet=3.7.4-1puppetlabs1 software-properties-common jiocloud-ssl-certificate run-one git
+
+#We are building puppet modules from source (ie run librarian-puppet on Puppetfile)
+#For librarian-puppet to run, we need the puppet-rjil source, hence we are installing puppet-jiocloud first
 if [ -n "${puppet_modules_source_repo}" ]; then
-  apt-get install -y git
+  apt-get install -y puppet-jiocloud
```
hkumarmk: if its building from source why you need to install this deb?

build_scripts/make_userdata.sh
hkumarmk: This should be a separate patch

hkumarmk:
@saneax In order to make this script work with vagrant why you need these changes?

I believe this script should work with vagrant much less changes. And you may only include those changes which are relevant to the purpose.

build_scripts/make_userdata.sh
saneax: I have not changed the logic from earlier, where it builds from source. So, Earlier, it would install all of this (below) and enter the block where we build by source. 
apt-get install -y puppet software-properties-common puppet-jiocloud jiocloud-ssl-certificate
if [ -n "${python_jiocloud_source_repo}" ]; then

to keep things similar for the source build I have kept it that way. for vagrant we absolutely dont need this, since installing will overwrite your poppet working dir.

build_scripts/make_userdata.sh
saneax: tests are failing without this going in, for the vagrant part

build_scripts/make_userdata.sh
hkumarmk: Can you please make only relevant change to make it work with vagrant and split other stuffs like cosmetic changes and other fixes as separate patches. That would be easier to review and to get it merged.

build_scripts/make_userdata.sh
sorenh: That won't work. run-one uses a semaphore named based on the command being run. From cron, the command is maybe-upgrade.sh. Here, it's "puppet apply --detailed-exitcodes etc.etc." Hence, run-one will happily run both maybe-upgrade as well as puppet apply here at the same time.

sorenh: Perhaps we could just use "flock -xn" without using run-one.

saneax: for #407, I will submit another patch, will help segregate it as per hkumar's suggestion.

build_scripts/make_userdata.sh
```diff
@@ -38,7 +38,12 @@ then
        dpkg -i internal.deb
 fi
 apt-get update
-apt-get install -y puppet software-properties-common puppet-jiocloud jiocloud-ssl-certificate
+apt-get install -y puppet software-properties-common jiocloud-ssl-certificate
+#we do not need puppet-jiocloud in vagrant images as they are mounting the puppet folders.
+if [ "${cloud_provider}" != "vagrant-box" ]
```
hkumarmk: I believe you should mount the folders from vagrant  under *.override. otherwise this is not going to work as the code below to this will set the module, hiera path to appropriate .overrides directories.

saneax: if we mount the below with overrirdes  (and then run userdata, we are entering into a different problem)
     config.vm.synced_folder("hiera/", '/etc/puppet/hiera.overrides/')                                                           
     config.vm.synced_folder("modules/", '/etc/puppet/modules.overrides/')                                                       
     config.vm.synced_folder("manifests/", '/etc/puppet/modules.overrides/rjil/manifests/')                                      
     config.vm.synced_folder("files/", '/etc/puppet/modules.overrides/rjil/files/')                                              
     config.vm.synced_folder("templates/", '/etc/puppet/modules.overrides/rjil/templates/')                                      
     config.vm.synced_folder("lib/", '/etc/puppet/modules.overrides/rjil/lib/')                                                  
     config.vm.synced_folder(".", "/etc/puppet/manifests.overrides")

so, in userdata script, the flow is like this..
install puppet-jiocloud (which will overwrite /etc/puppet, and which is the problem we are trying to address.)

when userdata runs, you run it without the option ($python_jiocloud_source_repo) so that puppet does not rebuild from git (this is so, to save time as well). 

so, now it enters the else part
where we are doing the puppet ini_settings

puppet apply -e "ini_setting { default_manifest: path => \"/etc/puppet/puppet.conf\", section => main, setting => default_manifest, value => \"/etc/puppet/manifests/site.pp\" }"

which is wrong again, as we have overriden paths in vagrant now.

so.. the userdata.txt says.. we are installing package puppet-jiocloud, for general.. and if ${python_jiocloud_source_repo} is populated we build from source and change the puppet.conf with the overrides.. if not, we assume puppet-jiocloud is installed and go with default path..

so, here are the problem parts on userdata..
1. install of puppet-jiocloud
2. running of puppet ini_settings 
so, we need to have a block for cloud kind of infra and an else Vagrant kind of block in userdata.txt similarly to the first patch I had submitted

PR #414, opened 2015-02-12
Setup mysql binary file replication

hkumarmk:
This is ready for testing - worked on my vagrant.

* Updated rjil::db class to configure mysql master server
* Updated the name of ocdb to ocdbm (master) and ocdbs (slave)
* Connect to master.mysql.service.consul
* Setup nightly mysql backup

This depends on PR jiocloud/puppet-orchestration_utils#7


hkumarmk:
please test this

manifests/db.pp
```diff
@@ -8,13 +8,50 @@
...
+  $mysql_data_disk           = undef,
+  $dbs                       = {},
+  $bind_address              = '0.0.0.0',
+  $server_id                 = undef,
+  $log_bin                   = 'mysql-bin.log',
+  $is_master                 = false,
+  $binlog_format             = 'mixed',
+  $sync_binlog               = 1,
+  $relay_log                 = 'mysql-relay-bin.log',
+  $repl_user                 = 'repl',
```
bodepd: this needs to be moved to secrets/common.yaml

hkumarmk: sure... i just added a default user/pass :)

hkumarmk:
please test this

lib/facter/openstack_mysql_leader.rb
```diff
@@ -0,0 +1,39 @@
...
+    setcode do
+      if leader
+        hostname
+      else
+        leader_name
+      end
+    end
+  end
+
+  Facter.add(:orc_node_status) do
```
bodepd: this fact is kind of confusing. Is leader a role specific or a node specific thing?

hkumarmk: oops I am rewriting the entire logic of fact with the logic here

https://github.com/JioCloud/jiocloud-specs/pull/11#discussion_r24721437

hkumarmk:
please test this

manifests/db.pp
hkumarmk: this is done

hkumarmk:
travis failure will be stopped after we merge the patch on orchestration_utils

manifests/db.pp
```diff
@@ -7,14 +7,83 @@
...
+        'max_connections' => $mysql_max_connections,
+        'datadir'         => $mysql_datadir,
+        'bind-address'    => $bind_address,
+        'server-id'       => $server_id_orig,
+        'log_bin'         => $log_bin,
+        'binlog_format'   => $binlog_format,
+        'sync_binlog'     => $sync_binlog,
+      }
+
+  if $is_master {
```
bodepd: what's the difference between master and leader? I assume master is for bootstrapping only?

bodepd: inline comments here would be very helpful

manifests/db.pp
```diff
@@ -7,14 +7,83 @@
...
+    consul_kv_mysql_masterdata {'openstack':}
+    Mysql_user<||> -> Consul_kv_mysql_masterdata<||>
+    Mysql_grant<||> -> Consul_kv_mysql_masterdata<||>
+    Mysql_database<||> -> Consul_kv_mysql_masterdata<||>
+    Consul_kv_mysql_masterdata<||> -> Rjil::Jiocloud::Consul::Service<| title == 'mysql' |>
+    Class['rjil::db'] -> Rjil::Service_blocker<| title == 'master.mysql' |>
+  } else {
+    if $::leader {
+      $override_mysqld_options = {
+        'read_only' => true,
```
bodepd: I'm confused at why read_only is set to true when our non-master becomes a leader.

manifests/db.pp
```diff
@@ -7,14 +7,83 @@
...
+        'relay_log' => $relay_log,
+      }
+      $consul_tags = ['slave']
+
+      consul_kv_mysql_slave_update{'openstack':
+        repl_user => $repl_user,
+        repl_password => $repl_user,
```
bodepd: did you mean to set this to user?

bodepd: why are you writing the password?

manifests/db.pp
```diff
@@ -7,14 +7,83 @@
...
+    # Add master data to consul kv
+    ##
+    consul_kv_mysql_masterdata {'openstack':}
+    Mysql_user<||> -> Consul_kv_mysql_masterdata<||>
+    Mysql_grant<||> -> Consul_kv_mysql_masterdata<||>
+    Mysql_database<||> -> Consul_kv_mysql_masterdata<||>
+    Consul_kv_mysql_masterdata<||> -> Rjil::Jiocloud::Consul::Service<| title == 'mysql' |>
+    Class['rjil::db'] -> Rjil::Service_blocker<| title == 'master.mysql' |>
+  } else {
+    if $::leader {
```
bodepd: are you sure this isn't supposed to be if ! $leader ?

manifests/db.pp
```diff
@@ -123,13 +189,29 @@
     require    => Mysql_user["monitor@${user_address}"],
   }
 
+  ##
+  # This user should be anabled from all slave nodes, need to see how we add
+  #   the access to specific hosts, so enabling from all hosts for now.
+  ##
+  mysql_user { "${repl_user}@%":
```
bodepd: this looks like it can cause race conditions.

manifests/db.pp
hkumarmk: Sure ill add inline comment. leader is the fact and is_master is the hiera data we set based on cluster_role fact (if the machine's cluster role is leader is_master is set). The reason of is_master and not is_leader is that, we setup mysql master on leader node for now, and thus it make sense the name is_master. We can change it to is_leader also if it make confusion as master is may be to mysql only.

manifests/db.pp
hkumarmk: Nops this just make sure this block will only be executed if the leader fact is available - this fact will not be available on first puppet execution and thus I dont want the mysql slave code executed (in fact that will add unnecessary service blocker).

manifests/db.pp
hkumarmk: It must be repl_password, ill fix it. and I did not understand your second question.

manifests/db.pp
bodepd: that is definitely not obvious reading the code. I thought leader was a fact that returned true or false. If you read it, is says:

````
if we are the leader, do
````

This definitely needs to be refactored to make it more clear

File: manifests/db.pp (hunk 6 above)
Comment (hkumarmk) [inline manifests/db.pp]: Nops its just checking whether the $leader fact is there or not (leader fact will always have a value), the fact will not be set in the initial puppet run.

File: manifests/db.pp (hunk 7 above)
Comment (hkumarmk) [inline manifests/db.pp]: Why you think so? Could you please bit explain this?

File: manifests/db.pp (hunk 5 above)
Comment (bodepd) [inline manifests/db.pp]: why are we actually passing the username/password and not an arbitrary key? All nodes have access to this data via facter, and I don't want passwords to leak to other places.

Comment (hkumarmk) [inline manifests/db.pp]: the nodes doesnt have access to user credentials from fact, the fact only have binary file name  and  position. This user credentials need to be provided to mysql while configuring slave.

File: manifests/db.pp (hunk 4 above)
Comment (hkumarmk) [inline manifests/db.pp]: I can change the fact name to leader_node

File: manifests/db.pp (hunk 7 above)
Comment (hkumarmk) [inline manifests/db.pp]: In fact the mysql_user will just create mysql users in individual mysql nodes - they are not shared in any aspects.

Comment (bodepd) [inline manifests/db.pp]: you are right, Im wrong, I was incorrectly thinking it is shared state.

File: manifests/db.pp (hunk 6 above)
Comment (bodepd) [inline manifests/db.pp]: maybe if defined($::leader) is a little more descriptive as to what is happening

File: manifests/db.pp (hunk 5 above)
Comment (bodepd) [inline manifests/db.pp]: oh wow. that's complicated. I'm going to think about that :)

Comment (bodepd) [inline manifests/db.pp]: maybe the fact can lookup the data from hiera? hiera has a command line facility? Is that easier/crazy? I'm not the biggest fan of facts doing things that require auth. I need to review the facts...

File: manifests/db.pp (hunk 4 above)
Comment (bodepd) [inline manifests/db.pp]: yes!!! and also change it to if defined($::leader_node) and put a comment above it explaining that's it's seeing if the fact is defined.

**Comment by hkumarmk** [comment]
@bodepd Made suggested changes

File: manifests/db.pp (hunk 5 above)
Comment (hkumarmk) [inline manifests/db.pp]: @bodepd im bit confused here ... in fact I did not understand what kind of problem you are seeing there.... I initiated a discussion on slack on this...

File: spec/classes/cinder_spec.rb [hunk 8]
```diff
@@ -42,7 +42,8 @@
       should contain_file('/usr/lib/jiocloud/tests/service_checks/cinder.sh').with_content(
         /check_http -H 10\.1\.1\.1 -p 8776/
       )
-      should contain_Cinder_config('database/connection').that_requires('Rjil::Service_blocker[mysql]')
+      should
+contain_Cinder_config('database/connection').that_requires('Rjil::Service_blocker[master.mysql]')
```
bodepd: accidental newline?

bodepd:
I've done as much review as I can without building out a test environment. Once we get a way to build environment, I can validate and hopefully merge

spec/classes/cinder_spec.rb
hkumarmk: oh yeah... ill fix it

bodepd:
rebased as #414

PR #417, opened 2015-02-12
Haproxy do not need floating IP

hkumarmk:


hkumarmk:
please test this

bodepd:
this is true also for prod?

hkumarmk:
@bodepd yes it is, we dont have any floating IP on undercloud in production. We NAT public IP to the IP assigned to haproxy node[s]

PR #422, opened 2015-02-14
correcting tab spaces (for consistency) and converting tabs to spaces

saneax:


saneax:
No need to run test, as the only change is change of tabs to spaces

PR #423, opened 2015-02-14
Using make_userdata.sh common.

saneax:
we populate the same variable cloud_provider=vagrant-vbox and populate
new source for hiera data called hiera/data/cloud_provider/vagrant-vbox.yaml

This with contrast to pull request #413 has lesser changes to make_userdata.sh
I would like to change the while loop endlessly via a variable, to run once against endlessly in case of errors. Since Vagrant requires to check for errors easily.


saneax:
please test this

saneax:
FAIL: setUpClass (tempest.api.compute.volumes.test_volumes_list.VolumesTestJSON), is this expected?

saneax:
FAIL: setUpClass (tempest.api.compute.volumes.test_volumes_list.VolumesTestJSON), is this expected?

saneax:
please test this

build_scripts/make_userdata.sh
```diff
@@ -38,11 +38,24 @@ then
   dpkg -i internal.deb
 fi
 apt-get update
-apt-get install -y puppet software-properties-common puppet-jiocloud jiocloud-ssl-certificate
```
bodepd: why not just install puppet-jiocloud and set up overrides to vagrant mounted directories (I feel like @hkumarmk  already asked this question). I would prefer this if possible so that we can keep the shell script as consistent as possible between environments.

bodepd: you can disregard this comment, I think I get it. This is a special case where you want to override content, but not install form the VM using librarian-puppet

build_scripts/make_userdata.sh
```diff
@@ -38,11 +38,24 @@ then
...
+
+#if vagrant box, we are not interested in installing puppet-jiocloud
+if [ "${cloud_provider}" == "vagrant-vbox"]
+then
+  apt-get install -y puppet software-properties-common jiocloud-ssl-certificate
+  env="${environment}:-virtual-vbox"
```
bodepd: can we just set this in the Vagrantfile?

Vagrantfile
```diff
@@ -56,43 +58,14 @@ Vagrant.configure("2") do |config|
...
-        config.vm.provision 'shell', :inline =>
-        "test -e puppet.deb && exit 0; release=$(lsb_release -cs);http_proxy=#{ENV['http_proxy']} wget -O puppet.deb http://apt.puppetlabs.com/puppetlabs-release-${release}.deb;dpkg -i puppet.deb;apt-get update;apt-get install -y puppet-common=3.6.2-1puppetlabs1"
-      else
-        config.vm.provision 'shell', :inline =>
-        "test -e puppet.deb && exit 0; release=$(lsb_release -cs);wget -O puppet.deb http://apt.puppetlabs.com/puppetlabs-release-${release}.deb;dpkg -i puppet.deb;apt-get update;apt-get install -y puppet-common=3.6.2-1puppetlabs1"
-      end
       config.vm.provision 'shell', :inline =>
-      'puppet apply --detailed-exitcodes --debug -e "include rjil::jiocloud"; if [[ $? = 1 || $? = 4 || $? = 6 ]]; then apt-get update; puppet apply --detailed-exitcodes --debug -e "include rjil::jiocloud"; fi'
-
+      "export consul_discovery_token=#{ENV['consul_discovery_token']};export layout=full;export map=#{map};export cloud_provider=vagrant-vbox;export environment=#{environment};/etc/puppet/manifests/build_scripts/make_userdata.sh;bash -x ./userdata.txt"
```
bodepd: I think cloud provider should be configurable b/c they are different vagrant providers

saneax: we will do this.

build_scripts/make_userdata.sh
saneax: I think we are already populating $cloud_provider from Vagrantfile. Or I did not get.

bodepd: this is defaulting to vbox here. Can we just set it from the Vagrantfile and remove this code?

saneax: we will do this.

saneax:
please test this

build_scripts/make_userdata.sh
```diff
@@ -38,11 +38,22 @@ then
   dpkg -i internal.deb
 fi
 apt-get update
-apt-get install -y puppet software-properties-common puppet-jiocloud jiocloud-ssl-certificate
+
+#if vagrant box, we are not interested in installing puppet-jiocloud
+if [ "${cloud_provider}" == "virtualbox"]
```
bodepd: virtualbox does not match the hiera file provided. Also, I'm not sure if I like calling this cloud_provider virtualbox b/c this patch doesn't really have anything to do with whether virtualbox is used (it's vagrant specific, not vagrant plugin specific). Would you mind using env=vagrant.

saneax: I should have looked at this, have made changes, now so as env is passed with value vagrant-vbox.

saneax:
please test this

saneax:
please test this

saneax:
please test this

build_scripts/make_userdata.sh
```diff
@@ -75,12 +86,19 @@ if [ -n "${puppet_modules_source_repo}" ]; then
...
+#Populate Facter Variables
 sudo mkdir -p /etc/facter/facts.d
 echo 'consul_discovery_token='${consul_discovery_token} > /etc/facter/facts.d/consul.txt
 echo 'current_version='${BUILD_NUMBER} > /etc/facter/facts.d/current_version.txt
 echo 'env='${env} > /etc/facter/facts.d/env.txt
 echo 'cloud_provider='${cloud_provider} > /etc/facter/facts.d/cloud_provider.txt
-while true
+
+#Run Puppet endlessly until the puppet code run is successful
+#Passing a new variable maximum nos of this loop to be executed, default endlessly
```
bodepd: why is this a good idea? What is the use case for wanted to break execution here even if things are not working?

saneax: IN Vagrant provisioning the shell runs endlessly, and a ctrl+c would put a lot of data on screen without being able to debug. I would presume, on vagrant we can fail out after a certain no. of retries. Is it not that the cron also tries to fix the issue?

bodepd: this patch was pretty close to being merged without adding more features to it.

build_scripts/make_userdata.sh
```diff
@@ -38,11 +38,22 @@ then
   dpkg -i internal.deb
 fi
 apt-get update
-apt-get install -y puppet software-properties-common puppet-jiocloud jiocloud-ssl-certificate
+
+#if vagrant box, we are not interested in installing puppet-jiocloud
```
bodepd: several problems with this logic, it will break the lxc use case (is anyone still using that?) Perhaps we need an environment name that encompasses all of the vagrant providers?

PR #430, opened 2015-02-17
WIP Depends on pull request https://github.com/JioCloud/python-jiocloud/pull/43 - Added a class debug_timeout 

amar266:
The class will generate a ini file in /etc directory
with name of debug_timeout.ini . Added relevent data in hiera
Changed deploy.sh to run python -m jiocloud.orchestrate with --debug_hosts arg

### DIFF
--- a/manifests/debug_timeout.pp
+++ b/manifests/debug_timeout.pp
@@ -1,0 +1,14 @@
+#
+# Class rjil::debug_timeout
+#
+class rjil::debug_timeout(
+  $debug_files,
+) {
+  ini_setting { 'debug_timeout':
+    path    => '/etc/debug_timeout.ini',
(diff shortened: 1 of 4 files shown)


manifests/debug_timeout.pp
```diff
@@ -0,0 +1,14 @@
+#
+# Class rjil::debug_timeout
+#
+class rjil::debug_timeout(
+  $debug_files,
+) {
+  ini_setting { 'debug_timeout':
```
bodepd: why is this called debug_timeout? Isn't this just the filename?

amar266: Yup, there was no specific reason to name so, it was creating a ini file to be usable for debug process.

PR #431, opened 2015-02-18
Support for other openstack cloud

viggates:
This patch would make puppet-rjil to deploy cloud on any openstack
cloud.


saneax:
I would suggest to create a hiera file for the undercloud specific and cloud provider specific secrets from a local resource ie line 26 on common.sh. 
Also, any undercloud specific resource create, would presumably if using nova create should use the jiocloud.apply_resouces call so as to use a similar call as done with hp cloud. I think then we can just call deploy.sh from the jumphost.

PR #436, opened 2015-02-23
Allow passing in a URL to a tarball with cached git data

sorenh:


build_scripts/make_userdata.sh
```diff
@@ -59,6 +59,14 @@ if [ -n "${puppet_modules_source_repo}" ]; then
     git merge -m 'Merging Pull Request' test_${pull_request_id}
     popd
   fi
+  if [ -n "${module_git_cache}" ]
```
bodepd: a few things here.

* needs to be modules overrides
* need to change librarian puppet from install to update

bodepd: * also, it needs to be based on https://github.com/JioCloud/puppet-rjil/pull/435 to be testable

sorenh: I don't think it should be modules.overrides. librarian installs in modules.

sorenh: Sorry, I totally misread. You're right. I'll fix it.

bodepd: also #435 is not ready

bodepd:
will be tested with #437

PR #438, opened 2015-02-27
Move tempest provision from rjil::tempest class to ocdb

hkumarmk:
rjil::tempest class would be running on gcp1 and some/all of the openstack
resources are created on ocdb, so it make sense to run the provisioning on ocdb
itself.


bodepd:
unit test failures fixed by #439

PR #441, opened 2015-03-02
Create appropriate configuration files to provision resources

hkumarmk:
Provisioning tempest resources need neutron.conf which is not there in ocdb1. So
adding appropriate config files.

NOTE: The original fix should be openstackclient based providers for openstack
objects :(


bodepd:
this got merged as a part of #454

PR #442, opened 2015-03-02
Only setup tempest_admin on vagrant, at and gate only

hkumarmk:
Moved tempest_admin config from common.yaml so it will only be setup on specific
environments. The idea is to only have tempest_admin on vagrant, at and gate
only and may be on staging also - this user will not be on production.


bodepd:
what is the motivation for this change?

hkumarmk:
We are not going to run admin tests on production, so the user is not required there.

bodepd:
please test this

PR #443, opened 2015-03-02
Pin Puppet to mitigate Future Parser errors

alokjani:
Currently, the latest version of Puppet (3.7.x) gets installed from
Puppetlabs during the initial setup.
Then next version (Puppet 4) has several language changes alongwith
a new parser that is enabled by default.
I am seeing that the puppet-consul module throws an "Evaluation Error"
when tested w/ future parser.

This patch ensures that 3.x is used until all dependent modules are
fully future parser compatible.

### DIFF
--- a/build_scripts/make_userdata.sh
+++ b/build_scripts/make_userdata.sh
@@ -8,6 +8,7 @@ export LC_ALL=en_US.UTF-8
 export LANG=en_US.UTF-8
 export LANGUAGE=en_US.UTF-8
 export layout="${layout}"
+export PUPPET_VER="3.7.2-1puppetlabs1"
 release="\$(lsb_release -cs)"
 if [ -n "${git_protocol}" ]; then
   export git_protocol="${git_protocol}"
@@ -38,7 +39,7 @@ then
   dpkg -i internal.deb
 fi
 apt-get update
-apt-get install -y puppet software-properties-common puppet-jiocloud jiocloud-ssl-certificate
+apt-get install -y puppet=${PUPPET_VER} puppet-common=${PUPPET_VER} software-properties-common puppet-jiocloud jiocloud-ssl-certificate
 if [ -n "${python_jiocloud_source_repo}" ]; then
   apt-get install -y python-pip python-jiocloud python-dev libffi-dev libssl-dev git
   pip install -e "${python_jiocloud_source_repo}@${python_jiocloud_source_branch}#egg=jiocloud"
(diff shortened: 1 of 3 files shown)


build_scripts/make_userdata.sh
```diff
@@ -72,6 +73,7 @@ if [ -n "${puppet_modules_source_repo}" ]; then
   puppet apply -e "ini_setting { basemodulepath: path => \"/etc/puppet/puppet.conf\", section => main, setting => basemodulepath, value => \"/etc/puppet/modules.overrides:/etc/puppet/modules\" }"
   puppet apply -e "ini_setting { default_manifest: path => \"/etc/puppet/puppet.conf\", section => main, setting => default_manifest, value => \"/etc/puppet/manifests.overrides/site.pp\" }"
   puppet apply -e "ini_setting { disable_per_environment_manifest: path => \"/etc/puppet/puppet.conf\", section => main, setting => disable_per_environment_manifest, value => \"true\" }"
+  puppet apply -e "apt::pin { '00-puppet': explanation => 'Lets use 3.x for now', packages => 'puppet puppet-common', priority => '501', version  => '3.7.*', }"
```
bodepd: why is this defined both here and in the manifest?

alokjani: @bodepd 
Here its to ensure that the pinning happens at the very first time the system is bootstrapped and before maybe-upgrade.sh has a scheduled run (possible dist-upgrade L#48 files/maybe-upgrade.sh)

The presence in the manifest is of-course imp for the agent version required during the node's life-cycle & when we decide to upgrade.

bodepd: each of these Puppet runs adds several seconds to the overall build time, is there a way that we can set up these dependencies inside of the puppet code?

bodepd:
one other question, are we not pulling the puppet packages from our mirrored repo? Or more specifically, why are we not seeing this for all builds?

build_scripts/make_userdata.sh
alokjani: Okay. In that case, I'll keep just the entry in jiocloud.pp and not make any in maybe-upgrade.sh.
And probably replace the apt::pin define with just a simple File resource ?

alokjani:
Currently we are pulling puppet from Puppetlabs [1] and not from any internal mirror.
I came a/c  a few errors [2] that caused the puppet run to fail w/ future parser enabled (default with 4). We'll start seeing these once Puppet 4 is launched and is the default latest.

[1]
https://github.com/JioCloud/puppet-rjil/blob/master/build_scripts/make_userdata.sh#L27
[2]
Error: Evaluation Error: No matching entry for selector parameter with value '1' at /etc/puppet/modules.overrides/consul/manifests/params.pp:26:17 on node bootstrap1-testjenkins-puppet-rjil-gate-713.node.consul
Error: Evaluation Error: No matching entry for selector parameter with value '1' at /etc/puppet/modules.overrides/consul/manifests/params.pp:26:17 on node bootstrap1-testjenkins-puppet-rjil-gate-713.node.consul

bodepd:
@alokjani a patch for future parser fixes would be greatly appreciated.

bodepd:
I'm not convinced this is going to be a problem, but I still need to verify the exact details. If I understand correctly, puppet4 is going to be released as an entirely different type of package with backwards compat breakages and a different name: "puppet-agent", if this is true, it means that "puppet" will also be 3.7.x, I'm still trying to get confirmation on the list.

alokjani:
Yep, just went through the one for puppet4 and it installs as puppet-agent

PR #451, opened 2015-03-09
WIP: Adding jenkins build server

saneax:
This installs the jenkins server


hiera/data/env/vagrant-vbox.yaml
```diff
@@ -17,8 +17,8 @@ nova::compute::libvirt::libvirt_virt_type: qemu
 rjil::jiocloud::consul::service::interval: 120s
 
 rjil::neutron::contrail::public_cidr: 100.1.0.0/24
-rjil::system::accounts::active_users: [soren,bodepd,hkumar,jenkins,consul,pandeyop,jaspreet,vivek,ahmad,vaidy,himanshu,rohit,sanjayu,vigneshvar,alokjani,amar,ynshenoy,hanish]
```
bodepd: if you have conflicts on this user with vagrant, it means they will exist in all environments

saneax: removed jenkins user from other environments as well, as the jenkins module is populating this user

bodepd: what about non-jenkins roles?

saneax: It was happening so, that rjil::base was realizing after jenkins::plugins causing duplicate declaration. Had to make changes to localuser.pp so jenkins can be defined in hiera for all the other roles.

PR #452, opened 2015-03-10
WIP: Add missing rspec tests

alokjani:
This patch adds missing tests for
- rjil::system
- rjil::test::*

### DIFF
--- a/spec/classes/test/ceph_radosgw_spec.rb
+++ b/spec/classes/test/ceph_radosgw_spec.rb
@@ -1,0 +1,31 @@
+require 'spec_helper'
+
+describe 'rjil::test::ceph_radosgw' do
+  let :params do
+    {
+      'port'      => 6000,
+      'ssl'       => false,
+    }
+  end
+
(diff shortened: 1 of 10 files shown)


spec/classes/test/ceph_radosgw_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  context 'with defaults' do
+    it do 
+      should contain_package('python-swiftclient')
+    end
+
+    it do
+      should contain_file('/usr/lib/jiocloud/tests/ceph_radosgw.sh') \
+        .with_content(/swift upload/) \
```
bodepd: you can also just specify 'with' and pass it a hash or arguments, this is the usual method for this kind of syntax

alokjani: I came across this alternative in some of puppetlabs' works and felt it improved readability.

The rspec shorthand is noted though. thanks!

PR #453, opened 2015-03-10
Create single disk raid0 on hp storage boxes

hkumarmk:
This works like this:

1. Facter hp_unassigned_disks to detect all unassigned disks
2. rjil::ceph::osd to call hp_drive type with the above fact as parameter
3. hp_drive will accept drive id as parameter and make raid 0 on individual
disks and let them visible to operating system.
4. install hpacucli package

NOTE: currently hpacucli is not there in any of the repo we use, this package
need to be added before we can merge this code.


theanalyst:
as per our understanding you're making raid0  individually on each of the
25 disks so that they are visible to the external raid controller. Just to
be clear that overall disks do not do raid0 themselves
and ceph is replicating & partitioning again. Also doesn't the server
support some sort of JBOD mode
so that this can be avoided altogether.

hkumarmk:
making raid0 is required for hp systems as those systems will not let operating system see the disks without that. It will just expose individual disks as it is, so no extra raid created.

theanalyst:
OK as expected, :+1:

sorenh:
Yeah, "create a RAID0 for each drive" *is* these system's JBOD mode. Awkward, but that's how it is. :)

theanalyst:
On Tue, Mar 10, 2015 at 12:47 PM, Soren Hansen <notifications@github.com>
wrote:

> Yeah, "create a RAID0 for each drive" *is* these system's JBOD mode.
> Awkward, but that's how it is. :)
>
Yeah, Assumed that was the case, just wanted to make sure.

lib/puppet/provider/hp_drive/default.rb
```diff
@@ -0,0 +1,25 @@
...
+  def create
+    hpacucli('ctrl slot=1 create type=ld
+',"drives=#{resource[:name]}","raid=#{resource[:raid_level]}")
+  end
+
+  ##
+  # I dont know if we wanted to support delete Lun, it is bit complecated and
+  # destroy can cause dataloss.
+  ##
+  def destroy
```
bodepd: maybe just fail here instead of warn (since the operation isn't supported anyways)

bodepd:
code looks good. Is there a node that I can use for testing?

lib/puppet/provider/hp_drive/default.rb
hkumarmk: This can be done.

hkumarmk:
@bodepd you may just boot a storage machine in undercloud controller and try it there.

lib/puppet/provider/hp_drive/default.rb
hkumarmk: this is done

PR #461, opened 2015-03-16
Expose glance backend direct url

hkumarmk:
Exposing glance backend direct url will enable faster boot from volume as in
that case it is just a rbd clone rather than image download to cinder volume.


hkumarmk:
please test this

PR #472, opened 2015-03-18
Hiera changes to use keystone configuration in rjil::tempest::provision

hkumarmk:
This patch to do hiera lookup to keystone configuration to configure
rjil::tempest::provision, without this puppet run in production fails because of
auth_host mismatch.


hkumarmk:
please test this

hkumarmk:
merging this one as it passed the gate, this need to be reached production in order to have successful puppet run there in production

PR #473, opened 2015-03-18
Add allowed local disk in the manifests

hkumarmk:
This patch add the nova configuration max_local_block_devices to
rjil::nova::controller manifest so that the value can be overriden. It have
taken the default one which is 3 but will need to be overriden in production
where we need no local disks.


hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

bodepd:
why doesn't this patch add boot volumes?

PR #481, opened 2015-03-20
Disable boot from volume on environments other than jio

hkumarmk:
Disabling boot from volume on other environments as it only required in
jiocloud.


bodepd:
do we have flavors that support this? Won't this just get the default 2GB of data?

bodepd:
oh, nm, this does all envs except jio. Can we go back to boot from disk on jio (or at least have flavors that support it) until this issue is resolved?

PR #482, opened 2015-03-23
Make sure /var/lib/nova/instances have right permission

hkumarmk:
I have seen couple of compute node have the above directory with root ownership
which was causing vm boot failures as nova compute process did not have write
permissions.


bodepd:
@hkumarmk are we still seeing this issue?

PR #483, opened 2015-03-23
hiera.yaml should be at /etc/hiera.yaml

saneax:
not sure, but for me the hiera data does not load without this for me.


bodepd:
Everything works fine without this patch, I'm going to close it. It can be reopened if it becomes an issue again.

PR #489, opened 2015-03-24
Update CORS headers in API service apache configs

rohit-k:
This change corrects the CORS header to 'always set'
It adds another header to allow 'Content-Type' in the
request headers.
An explicit rewrite rule to return 200 for the OPTIONS
request has also been added because OpenStack services
(especially Keystone) yield 404 for OPTIONS requests.


rohit-k:
please test this

PR #492, opened 2015-03-24
minor fixes

alokjani:
- mode should be a 4 digit octal value
- reduce class not documented travis noise


bodepd:
please test this

bodepd:
I babysit this and watched it fail, it just took hours b/c of what I believe is slow disk write issues on some nodes and some DNS reliability issues.

PR #494, opened 2015-03-24
Updated description to setup Dev Env using Vagrant

mayankkapoor:
Updated readme.md with up to date steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,9 @@ verified until glance has registered (maybe this is actually not a problem...)
 
 # Development environment
 
+You can setup a Dev environment on a bare-metal node with enough resources using Vagrant. Once you have the bare-metal
+server available, follow these steps.
+
 ## Installing vagrant
 
 [Install Virtualbox](http://www.virtualbox.org/manual/ch01.html#intro-installing)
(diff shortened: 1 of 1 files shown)


saneax:
I thought we are deprecating the vagrant way? 
The vagrantfile and deploy.sh differ on the initial setup.

README.md
```diff
@@ -499,25 +501,20 @@ customize the roles to be populated by adjusting
...
+    vagrant up <role>
 
 #### vagrant provision
 
 To re-run Puppet provisioning (for interactive development of modules):
 
-````
-vagrant provision <role>
-````
+    vagrant provision <role>
```
bodepd: why did you change these from block quotes?

README.md
```diff
@@ -419,79 +422,78 @@ The default version of Vagrant in the Ubuntu 14.04 Repo is 1.4.3 which causes an
...
+    export consul_discovery_token=ca004..7f42f3
+
+#### 4. Install puppet modules (using sudo -E)
 
-* setup your extra environment variables
+First, you need to make sure that your Puppet module dependencies are installed. 
+NOTE: make sure that you install librarian-puppet-simple and *not* librarian-puppet!!!!
 
-The Vagrantfile accepts a few additional environment variables that can be used to further customize the environment.
+    gem install librarian-puppet-simple # or sudo -E gem install librarian-puppet-simple
```
bodepd: in general, it's preferred to not install libraries into system locations for user functions. I guess the reason that I don't mention sudo here is that I had already installed rvm in my home directory.

README.md
mayankkapoor: These weren't showing up as code formatting when I ran them through markdown dingus (http://daringfireball.net/projects/markdown/dingus). Changed to indented to show them as code snippets. If you have a better recommendation, let me know. Thanks.

bodepd:
like @saneax mentioned before, the general consensus was to deprecate the vagrant functionality for openstack deployments. It's very useful for iterative development, but the full openstack env is HUGE and only getting bigger (we have patches out that add 3 more nodes). It would be possible for developers to create custom layouts for iterative development on subcomponents of the stack (but I'm not sure if it will be worth it considering the learning curve of building these). As a side note, I still use it to work on small things (like the external layout, but I haven't used it to build out openstack in ages)

The idea was that we could provide similar functionality via some local dev env rsync commands to push code out to the builds from our prod environment. I've never heard anything about if these tools were written or not.

One of the other issues was that there are differences between the acceptance test env and vagrant builds that resulted in tons of errors b/c we were not gating the vagrant way of building things (so had no reliable way to know if patches effected it). There is a patch out there intended to make the environments more similar that is currently waiting for a response (#423) .

I also tried to roll it out as a service for developers to use and it was a nightmare, people were resource constrained, so they tried to support multiple dev envs on the same bare-metal machine (kind of an awful implementation of a cloud). Needless to say it was a disaster.

README.md
```diff
@@ -419,79 +422,78 @@ The default version of Vagrant in the Ubuntu 14.04 Repo is 1.4.3 which causes an
...
 
-We have designed a vagrant based solution that can be used to create VMs that
-very closely resemble the openstack environments that we are deploying.
+The following initial setup steps are required to use the vagrant environment:
 
-When spinning up VMs for local development, Vagrant/VBox would need to be run on the physical host that has VT-x  enabled in its BIOS. Currently, this setup cannot be provisioned and tested inside a VM itself (KVM/VBox).
+#### 1. Set a local system squid proxy to access the internet from the server
 
-### getting setup
+Write following lines at the bottom of .bashrc file
```
bodepd: these parameters would be problematic as general parameters b/c they are using a virtual network as opposed to a more sane system network if you wanted to use the proxy for all local operations. I usually create a local rc file in the puppet-rjil directory. It might also make sense to name them something that doesn't actually mean something on the system (like vagrant_http_proxy, etc)

README.md
```diff
@@ -419,79 +422,78 @@ The default version of Vagrant in the Ubuntu 14.04 Repo is 1.4.3 which causes an
...
+    export env=vagrant-lxc # used to customize the environment to use
 
-* clone project:
+Then do `$ source .bashrc`
 
-````
-git clone git://github.com/jiocloud/puppet-rjil
-````
+In order to use proxy with sudo command, use sudo with -E option, e.g., 
+    sudo -E apt-get update
```
bodepd: why are you doing this?

README.md
```diff
@@ -419,79 +422,78 @@ The default version of Vagrant in the Ubuntu 14.04 Repo is 1.4.3 which causes an
...
 
-* setup tokens (this is required for setting up consul)
+In order to use proxy for apt, e.g., create a file in /etc/apt/apt.conf.d/90_proxy. Write following lines:
+    Acquire::http::proxy "http://10.135.121.138:3128";
+    Acquire::https::proxy "https://10.135.121.138:3128";
```
bodepd: also, why are you doing this? Looks out of scope.

README.md
```diff
@@ -419,79 +422,78 @@ The default version of Vagrant in the Ubuntu 14.04 Repo is 1.4.3 which causes an
...
-gem install librarian-puppet-simple
-# from root level of this repo
-librarian-puppet install
-````
+    source newtokens.sh
 
-* create rjil directory in modules
-This step is required b/c the top level directory (ie: puppet-rjil is a module itself and
-it needs to exist in the modulepath for the vagrant environment.
+You'll see a consul ID. Copy that consul ID and paste onto last time of ~/.bashrc
```
bodepd: there may be problems with reusing the same id, you need to be careful that you always recreate the env from scratch everytime, or old machines will join the new cluster.

README.md
bodepd: weird, if you click the view button, they both render correctly. I don't know why that is :)

mayankkapoor: Yes on Github they both render correctly. Found the reason: Github uses a special flavor of markdown with "fenced code blocks" https://help.github.com/articles/github-flavored-markdown/#fenced-code-blocks. So both should work. I like using the web dingus on daringfireball.net to show me how the final markdown looks like. In the future I'm not going to change these code blocks.

README.md
mayankkapoor: We had given the proxy settings in the rc file of the local user. To export this settings in sudo, we're using sudo -E.

README.md
mayankkapoor: We wanted to provide these proxy settings to apt. The two settings (sudo -E and proxy for apt) are independent and can be used exclusively (only one is needed for apt).

README.md
mayankkapoor: Updated this with the warning.

README.md
mayankkapoor: Updated with both options (incl. disadvantages of each approach).

PR #496, opened 2015-03-24
Use http proxy node as dns server

hkumarmk:
Use http_proxy node as dns server to other nodes.


hkumarmk:
please test this

hkumarmk:
please test this

bodepd:
I'm very curious what your motivation is here. I saw some issues where VMs booted today and did not have access to DNS. Is that the issue you observed that resulting in this patch?

PR #507, opened 2015-03-30
Combined multiple puppet runs to one

hkumarmk:
This should reduce couple of minutes


bodepd:
not sure about a couple minutes, but I would believe 30 seconds :)

bodepd:
please test this

bodepd:
this failed in a way that I have never seen before. DB migrations never ran b/c the mysql service was not correctly functioning when they ran.

This is related to an old ticket #85. @sorenh you mentioned that you were going to see if you could find a way to programmatically make a determination if we should run the db_migration scripts? That lack of retry on that action is what caused this issue.

bodepd:
please test this

bodepd:
please test this

Issue #444, opened 2015-03-02
unused variable $fail


bodepd:
rjil::ceph::mon::mon_config has a referenced variable not set anywhere.

https://github.com/JioCloud/puppet-rjil/blob/master/manifests/ceph/mon/mon_config.pp#L50

alokjani:
Fixed in #510

PR #510, opened 2015-03-31
Set $fail within local scope for Ceph mons

alokjani:
This patch fixes the `top-scope variable being used without an
explicit namespace` warning by setting $fail to true for
manifests/ceph/mon/mon_config.pp
Also set $fail explicitly in manifests/ceph/mon_config.pp

This closes issue #444

### DIFF
--- a/manifests/ceph/mon/mon_config.pp
+++ b/manifests/ceph/mon/mon_config.pp
@@ -52,7 +52,7 @@ class rjil::ceph::mon::mon_config (
     ::ceph::conf::mon_config{ $other_mons: }
   } elsif ! $leader {
     runtime_fail {'monlist_empty_fail':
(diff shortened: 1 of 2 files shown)


manifests/ceph/mon/mon_config.pp
```diff
@@ -51,6 +51,7 @@
   if ! empty($other_mons) {
     ::ceph::conf::mon_config{ $other_mons: }
   } elsif ! $leader {
+    $fail = true
```
bodepd: you can just set this directly in the resource since it's not used anywhere else.

alokjani: done.

bodepd:
please test this

PR #491, opened 2015-03-24
Squid proxy changes

hkumarmk:
Changed squid proxy setting according to squid-deb-proxy changes -
http://bazaar.launchpad.net/~mvo/+junk/squid-deb-proxy/view/head:/squid-deb-proxy.conf

PR #511, opened 2015-04-01
WIP: Make the httpproxy role configurable

alokjani:
This patch is to make httpproxy role configurable from hiera.


hkumarmk:
This change is already there in #491

alokjani:
Looks like i missed that one. I'll close this PR.

PR #520, opened 2015-04-07
Override domain name in dhclient

hkumarmk:
This is to configure dhclient.conf to override domain-name from dhcp server.


hkumarmk:
please test this

hkumarmk:
please test this

Issue #521, opened 2015-04-07
contrail intermittent issues


hkumarmk:
Sometimes while initializing (first time startup) contrail-api, I see below issues.

It seems contrail is trying to create the object which are already exists in zk. I see this is happening for multiple objects one after other. 

I have single node contrail node using opencontrail 1.2.

## contrail-api trying to create an object in zk, and throw nodeexists error.

INFO:api-0:Sending request(xid=1): Create(path=u'/fq-name-to-uuid/global_system_config:default-global-system-config', data='19f9ae74-f233-4ff8-b030-e2582ffab468', acl=[ACL(perms=31, acl_list=['ALL'], id=Id(scheme='world', id='anyone'))], flags=0)
INFO:api-0:Received error(xid=1) NodeExistsError((), {})

## The object api is trying to create is already exist in zk. 
[zk: localhost:2181(CONNECTED) 1] get /fq-name-to-uuid/global_system_config:default-global-system-config
12cf3860-d944-4fbf-8604-6fe6545a12a8
cZxid = 0x7
ctime = Tue Apr 07 04:03:54 UTC 2015
mZxid = 0x7
mtime = Tue Apr 07 04:03:54 UTC 2015
pZxid = 0x7
cversion = 0
dataVersion = 0
aclVersion = 0
ephemeralOwner = 0x0
dataLength = 36
numChildren = 0

## I see below errors in ifmap server, I dont know if that might be the issue

2015-04-07 05:17:03,600 [main] INFO   - ChannelAcceptor: Listening on port 8444 for incoming certificate-based authentication connections
2015-04-07 05:17:03,600 [main] INFO   - irond is running :-)
2015-04-07 05:17:07,076 [pool-6-thread-2] ERROR  - ChannelThread: Receiving request failed
2015-04-07 05:17:07,076 [pool-6-thread-2] ERROR  - ChannelThread: SSLHandshakeException: Client doesn't know about our certificate (?)
2015-04-07 05:17:07,076 [pool-6-thread-2] ERROR  - ChannelThread: Setting channel 127.0.0.1:40616:1 into state 'BROKEN'
2015-04-07 05:17:10,093 [pool-6-thread-2] ERROR  - ChannelThread: Receiving request failed
2015-04-07 05:17:10,093 [pool-6-thread-2] ERROR  - ChannelThread: SSLHandshakeException: Client doesn't know about our certificate (?)
[...]

PR #523, opened 2015-04-07
Casandra and zookeeper should be up before contrail packages

hkumarmk:
Created a service blockers for some packages which start services to use the
cassandra and zookeeper, so that they will not be installed unless cassandra and
zookeeper are started. The packae installation can start the application and the
default configuration may be enough to write the data to either of these
application.

We have seen a possible situation when zookeeper is up, and contrail-api is
starting but cassandra is not writable there, in which case, contrail-api can
create some zookeeper entries while creating cassandra entries, but since
cassandra is not writable, that operation fail (writing cassandra). So whenever
they try this operation again, the code try to create the uids using zk, but
since zk have those object already exists, the operation fail and thus contrail
api fail (this must be a bug in contrail api though).

Please see https://github.com/JioCloud/puppet-rjil/issues/521 for more details

### DIFF
--- a/manifests/cassandra.pp
+++ b/manifests/cassandra.pp
@@ -55,4 +55,13 @@ class rjil::cassandra (
     require           => Host['localhost'],
   }
 
+  ##
+  # I dont know if the port check is enough here - refer
+  # https://github.com/JioCloud/puppet-rjil/issues/521 for more details
+  ##
+  rjil::jiocloud::consul::service { "cassandra":
+    tags          => ['real','contrail'],
+    check_command => '/usr/lib/jiocloud/tests/check_cassandra.sh',
+  }
+
 }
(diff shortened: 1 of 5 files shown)


hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

bodepd:
Can we fix the packages not to start the service?

bodepd:
test failed with the same symptom that this patch was intended to fix.

hkumarmk:
The problem it is trying to solve here would have solved by #524

bodepd:
oh good :) can we close this, b/c I am also kind of solving this in a totally different way for the scaling patch :)

PR #540, opened 2015-04-16
Fix: this will fix the issues on contrail rt creation for the fip pool

hkumarmk:


hkumarmk:
@sorenh @bodepd it is a tiny patch to fix the issues with the contrail_rt addition we faced yesterday in production. Could you please merge it real quick?

PR #541, opened 2015-04-16
fixed dependency errors

hkumarmk:


hkumarmk:
please test this

bodepd:
what is the symptom of this?

hkumarmk:
bodepd, i m still working on this patch, - this was because we had changed the argument for contrail_rt, but missed to change the name of deps, but it need more changes (we need to add hiera for rt number so that it support different rts)

hkumarmk:
please test this

PR #544, opened 2015-04-17
Parameterized haproxy configs

hkumarmk:
This will make the configuration changes to be easier.

Also increased default timeout values to 20sec.

### DIFF
--- a/manifests/haproxy.pp
+++ b/manifests/haproxy.pp
@@ -5,37 +5,52 @@
 #
 #
 class rjil::haproxy (
-  $consul_service_tags = []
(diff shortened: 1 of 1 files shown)


hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
this test is actually passed, because of couple of test failures which are because of other reasons

1. the osd test need to be bit more changed to make it work with loop device 
2. vgw issue which are filed as issue

hkumarmk:
@bodepd @sorenh could you please review it?

PR #546, opened 2015-04-17
Avoid finding /dev/sdaa when looking for /dev/sda

sorenh:


hkumarmk:
Since this is very small change and gate test is not going to test this change (we only have two disks here, and it only can be tested if we have disks with name sdaa or more. So im just merging this

PR #547, opened 2015-04-17
osd test should handle loop devices also

hkumarmk:
This patch will make ceph test work for loop devices also where the partition
have different naming convention.

### DIFF
--- a/templates/tests/ceph_osd.sh.rb
+++ b/templates/tests/ceph_osd.sh.rb
@@ -7,5 +7,9 @@ function fail {
   exit 2
 }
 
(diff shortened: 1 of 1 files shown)


hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

sorenh:
Just FYI, I have a patch planned for Overcast that will let us spin up st nodes with as many volumes attached as we want so that we don't have to muck around with loop devices.

hkumarmk:
sure

PR #549, opened 2015-04-17
Use eth0 by default

sorenh:


sorenh:
These values are overridden in the gate, so testing them is pointless.

PR #542, opened 2015-04-16
Collectd with consul-alerts

amar266:
This Class will Install consul-alerts and enable slack
for notifications.

PR #555, opened 2015-04-20
<WIP> Testing Collectd

amar266:


amar266:
Please test this

amar266:
please test this

amar266:
please test this

bodepd:
to paraphrase soren based on a conversation that we had.

This patch has a single generic consul service to see if any any of the defined thresholds are failing and the python script generically just pushes all types of failures to this service without actually having to understand the structure of the alerts coming from collectd. This is fine for a single alert type, but when we have multiple types, all of those events will feed into a single service. For better visibility for each type of alert, it makes sense to have a consul service per alert.

In order to implement this, you need something like the following:

1. you have to know the actual schema of thresholds in collectd. This is defined in https://collectd.org/documentation/manpages/collectd-threshold.5.shtml as:

````
There are basically two types of configuration statements: The Host, Plugin, and Type blocks select the value for which a threshold should be configured. The Plugin and Type blocks may be specified further using the Instance option. You can combine the block by nesting the blocks, though they must be nested in the above order, i.e. Host may contain either Plugin and Type blocks, Plugin may only contain Type blocks and Type may not contain other blocks. If multiple blocks apply to the same value the most specific block is used.
````

2. create a public defined resource that can create both thresholds as well as associated consul services
Given the above definition, it should be easy enough to create a class that can both build collect config and associated consul services

3. update python script so that when it finds an alert, it can figure out what service check to forward it to

@sorenh please comment to make sure that this captures your thoughts from before

amar266:
please test this

amar266:
Merged all changes to pull request https://github.com/JioCloud/puppet-rjil/pull/542.

PR #562, opened 2015-04-21
Use eatmydata for apt-get and dpkg for at, to speedup the things

hkumarmk:
This patch eatmydata for apt-get and dpkg in at environment. This should speedup
package installation and thus overall gate/at execution time.


hkumarmk:
please test this

bodepd:
I tried this before on hp cloud and the results were not significant. I'm curious to see if you have different results :)

hkumarmk:
okay, I dont know I just thought to try this as @sorenh is not thinking we could reliably make the golden image based deployment working

hkumarmk:
please test this

sorenh:
please test this

PR #570, opened 2015-04-27
adding logrotate

anshprat:


anshprat:
Please test this

manifests/jiocloud/logrotate.pp
```diff
@@ -0,0 +1,13 @@
...
+    logrotate:
+    ensure => installed
+  }
+  file {
+    '/etc/logrotate.conf':
+      ensure  => file,
+      mode    => '0444',
+      source  => 'puppet:///modules/rjil/logrotate-default.conf';
+  }
+}
```
sorenh: The formatting here is a bit off.

Make it:

    package { logrotate:
      ensure => installed,
    }

    file { '/etc/logrotate.conf':
      ...
    }

manifests/jiocloud/logrotate.pp
```diff
@@ -0,0 +1,13 @@
+class rjil::jiocloud::logrotate {
+  include logrotate
```
sorenh: I'm a bit confused what this logrotate module does.... If it doesn't even install the package or provide a default config, what does it do, and why are we including it?

manifests/jiocloud/logrotate.pp
anshprat: done

manifests/jiocloud/logrotate.pp
```diff
@@ -0,0 +1,11 @@
...
+  package { logrotate:
```
hkumarmk: you dont need package and file, the module is doing all these for you.

manifests/jiocloud/logrotate.pp
```diff
@@ -0,0 +1,11 @@
...
+    ensure => installed
+  }
+  file { '/etc/logrotate.conf':
```
hkumarmk: Rather than the hardcoded file, you would need logrotate::rule called in appropriate places - e.g logrotate::rule{glance-api: } and logrotate::rule{glance-registry:} on rjil::glance

anshprat: Yes, that was the starting idea, but I couldn't get the module to play nice with the rjil class for the default config. When I tried logrotate::base for the default config, I got the error stating logrotate::base class not found / does not exist (similar lines, dont remember the exact error).
Basically at this point of time, I am trying just to edit the default conf file rather than edit all files. Suggestions?

hkumarmk: where did you used logrotate::base? i see it should work without issues. if you could make the code snippet/in this PR, I could help you.

anshprat: This is what I tried:

anshup@ubuntu:~/git/ril/puppet-rjil$ cat manifests/jiocloud/logrotate.pp 
class rjil::jiocloud::logrotate {
  include logrotate
  logrotate::base { "Default conf file":
    source => 'puppet:///modules/rjil/logrotate-default.conf';
  }
}

This is the error I get:

Error: Puppet::Parser::AST::Resource failed with error ArgumentError: Invalid resource type logrotate::base at /home/anshup/git/ril/puppet-rjil/modules/rjil/manifests/jiocloud/logrotate.pp:5 on node ubuntu.handwa.sh
Error: Puppet::Parser::AST::Resource failed with error ArgumentError: Invalid resource type logrotate::base at /home/anshup/git/ril/puppet-rjil/modules/rjil/manifests/jiocloud/logrotate.pp:5 on node ubuntu.handwa.sh

anshprat:
I think I have come across a better solution, closing this, will open a new one tomorrow.

manifests/jiocloud/logrotate.pp
hkumarmk: The reason you are getting this error because you are calling logrotate under rjil::jiocloud::logrotate without providing inner logrotate in top namespace, without this, puppet is trying to load rjil::jiocloud::logrotate::base which will be failed.


You dont need rjil::jiocloud::logrotate, you just need to add below code under rjil::base

include ::logrotate::base

Then call appropriate ::logrotate::rule under specific classes as I mentioned above.

PR #571, opened 2015-04-27
Logrotate

anshprat:
addng logrotate - This should be better than the last PR


anshprat:
This has dependency on https://github.com/JioCloud/puppet-logrotate/pull/1/files

anshprat:
Please test this

Puppetfile
```diff
@@ -195,3 +195,7 @@ mod 'thias/squid3',
 mod 'puppetlabs/dhcp',
   :git => "#{base_url}/JioCloud/puppetlabs-dhcp",
   :ref => 'origin/master'
+
+mod 'Jiocloud/logrotate',
```
bodepd: can you change this name to indicate the upstream? I'd like for it to be really easy to figure out from this file what things we have forked (and from where)

anshprat: done

hkumarmk:
why you need to fork the module? you are doing the thing you have done in puppet-rjil (hardcoding the config file) in the module, which is not necessary.

anshprat:
The logrotate::rule is for respective files under /etc/logrotate.conf.d . What I want to edit right now is only the /etc/logrotate.conf to change the default and not the respective files under logrotate.conf.d.

The default file is how the module is providing the default config, hence just editing it presently to provide the defaults we need. 

Let me know if we should rather add all logrotate configs to puppet. It will be more work and has higher entropy and will need to check all components and environment. Just editing the default file is something that fits the bill presently.

hkumarmk:
I would say that is the way to go, rather than making logrotate.conf for everything and update it for everything, you create rule part of the component deployment, so you will not miss the configs,

anshprat:
being done via PR 574

PR #574, opened 2015-04-28
Logrotate cinder

anshprat:
specimen for logrotate inclusion code review


anshprat:
Please test this

anshprat:
Please test this

manifests/cinder.pp
```diff
@@ -260,5 +260,9 @@
   }
 
   rjil::jiocloud::consul::service { 'cinder-backup': }
-
+  
+  class  { 'rjil::jiocloud::logrotate':
```
hkumarmk: This should not be a class, because in case of a classs, you cannot reuse it. In fact you dont need anything under rjil::jiocloud, you can just directly call logrotate::rule from here (and other classess).

anshprat: Hmm, I am infact trying to make a reusable snippet (see manifests/jiocloud/logrotate.pp in this commit) (Since most of the logrotate configs are gonna be similar). Else will end up making same calls to ::rule with so many confs. So guess I should be looking at define?

hkumarmk: Sure, yes, you should do a define then. Also i dont see 'include logrotate' make any changes there, - even without is things should work

Puppetfile
```diff
@@ -195,3 +195,7 @@ mod 'thias/squid3',
 mod 'puppetlabs/dhcp',
   :git => "#{base_url}/JioCloud/puppetlabs-dhcp",
   :ref => 'origin/master'
+
+mod 'rodjek/logrotate',
+  :git => "#{base_url}/JioCloud/puppet-logrotate/",
+  :ref => 'origin/master'
```
sorenh: You need to also add this in .fixtures.yml. There are two different systems handling these dependencies: One for during testing and another for runtime. I know... What could possibly go wrong?

manifests/jiocloud/logrotate.pp
```diff
@@ -0,0 +1,18 @@
+define rjil::jiocloud::logrotate(
+  $service,
+  $logfile,
+  $rotate_every = 'daily',
+  $rotate = 60,
+  $compress = true,
+  $delaycompress = true,
+  $ifempty = false,
```
sorenh: Please align all the ='s.

manifests/jiocloud/logrotate.pp
```diff
@@ -0,0 +1,18 @@
...
+) {
+  logrotate::rule{$service:
```
sorenh: Space between { and $service, please.

manifests/jiocloud/logrotate.pp
```diff
@@ -0,0 +1,18 @@
...
+    path => $logfile,
+    rotate => $rotate,
+    rotate_every => $rotate_every,
+    compress => $compress,
+    delaycompress => $delaycompress,
+    ifempty => $ifempty
```
sorenh: Please align all the =>'s.

manifests/jiocloud/logrotate.pp
```diff
@@ -0,0 +1,18 @@
...
+    }
+  }
```
sorenh: A few style comments:

The closing brace should line up with the word whose block it closes. So, the } in line 17 should be in the third column to align with "logrotate::rule" and the final closing brace in line 18 should be in the first column, aligning with "define".

There should be a spce between { and $service in line 18

anshprat:
yay, .fixtures.yml did the trick! :+1:

anshprat:
Please test this

manifests/cinder.pp
```diff
@@ -260,5 +260,17 @@
...
+    logfile => '/var/log/cinder/cinder-api.log'
+  }
+  rjil::jiocloud::logrotate { 'cinder-schedular':
+    service => 'cinder-schedule',
+    logfile => '/var/log/cinder/cinder-schedular.log'
+  }
+  rjil::jiocloud::logrotate { 'cinder-volume':
+    service => 'cinder-volume',
+    logfile => '/var/log/cinder/cinder-volume.log'
+  }
```
anshprat: These multiple calls may not be the most elegant way of doing this. However all attempts to use array iteration failed due to use of define within define (logrotate::rule is a  define provided by module being used within rjil::jiocloud::logrotate define)

```|
$cinder2_logs = [ 'cinder-api', 'cinder-scheduler', 'cinder-volume', 'cinder-backup']
rjil::jiocloud::logrotate { $cinder2_logs:
  service     => $cinder2_logs,
  logfile     => $name
}
```


above fails with duplicate declaration error :!


.  Without the rjil::jc::logrotate define, will end up using classes for each modules to use - example

```|
class cinder_logs {
  $cinder_logs = [  'cinder-scheduler', 'cinder-volume', 'cinder-backup']
 
  logrotate::rule{ $cinder_logs:
    path          => "/var/log/cinder/${cinder_logs}.log",
    rotate        => 60,
    rotate_every  => daily,
    compress      => true,
    delaycompress => true,
    ifempty       => false
  }
```
 
Any other ideas? If not, then is it good to be merged? Would like to proceed per component rather than all components together (some of the logrotates have extensive pre/post scripts as part of defaullt logrotate which might take some time to be done correctly in puppet).

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

Puppetfile
```diff
@@ -195,3 +195,7 @@ mod 'thias/squid3',
 mod 'puppetlabs/dhcp',
   :git => "#{base_url}/JioCloud/puppetlabs-dhcp",
   :ref => 'origin/master'
+
+mod 'rodjek/logrotate',
+  :git => "#{base_url}/JioCloud/puppet-logrotate/",
```
hkumarmk: why do we need this fork?

anshprat: we dont, it was there so I used it

anshprat: though there is one change that is there on upstream master, which will be nice to have in our fork master, (the tagged version in upstream does not have this change)

anshprat: this https://github.com/rodjek/puppet-logrotate/commit/89ee645e2350045be48df7ac3ef86cbe4cd9b096

manifests/cinder.pp
hkumarmk: This will not work, either you need to make your define to use some defaults - service to defaults to $name, and logfile to default to /var/log/${name}.log. If not, you have to define all of them separately as you did right now.

Puppetfile
hkumarmk: you can use rodjek/ master rather than tagged one

hkumarmk:
dont you need to add logrotate::base?

hkumarmk:
I would suggest to add logrotate::base in rjil::base, so it will be there on all nodes, and then add the rules to appropriate application deployment code.

manifests/jiocloud/logrotate.pp
```diff
@@ -0,0 +1,18 @@
+define rjil::jiocloud::logrotate(
+  $service,
+  $logfile,
+  $rotate_every  = 'daily',
+  $rotate        = 60,
+  $compress      = true,
+  $delaycompress = true,
+  $ifempty       = false,
+) {
+  logrotate::rule{ $service:
```
hkumarmk: you can add "include logrotate::base" here also

anshprat:
hmm, well, logrotate::base is not required. logrotate::rule
from the docs - 
The only thing you'll need to deal with, this type configures a logrotate rule. Using this type will automatically include a private class that will install and configure logrotate for you.

so as long as any ::rule is defined we are good. though logrotate::base will take care of the basic problem we had (defaults being overridden everywhere). In fact, using ::base with the patched default takes care of the problem at hand everywhere.

manifests/jiocloud/logrotate.pp
anshprat: https://github.com/rodjek/puppet-logrotate/blob/master/manifests/rule.pp#L404  - including base with rule would be redundant

hkumarmk:
It is better you have it, that will make sure the entire logrotate is in good shape, and logrotate.conf also is managed by puppet - what if the line logrotate.d is commented somehow?

hkumarmk:
oh okay that make sense as ::rule have it included.

hkumarmk:
So as soon as you change the module reference to upstream rather than jiocloud and test passed, I am good to get this merged (You may also have to make ssure logrotate works as expected).

anshprat:
upstream master or upstream tag? coz upstream tag does not have the btmp patch. and if we can use our fork, with the patched defaults, we can fix our basic problem faster, else will need to wait till every conf is puppetized

hkumarmk:
that patch is already merged upstream, so you may use upstream master.

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
@hkumarmk @sorenh @bodepd guess this is good to be merged now? Please check.

anshprat:
please test this

anshprat:
please test this

PR #584, opened 2015-05-05
Logrotate_combo

anshprat:
Will be combining more of logrotate for various modules under this.


anshprat:
please test this

anshprat:
please test this

bodepd:
one comment here and I know that @sorenh doesn't always agree with me on this, but for such a simple change-set, it is not required to have so many commits. I like the idea of having as few commits as possible so as to keep history usable.

of coarse, the disadvantage to rebasing is that it's harder to understand how you are making changes once you start getting review. I think a reasonable compromise is to clean up history before you start soliciting feedback, then issue new comments in response to requests to change things.

Puppetfile
```diff
@@ -195,3 +195,7 @@ mod 'thias/squid3',
 mod 'puppetlabs/dhcp',
   :git => "#{base_url}/JioCloud/puppetlabs-dhcp",
   :ref => 'origin/master'
+
+mod 'rodjek/logrotate',
+  :git => "#{base_url}/rodjek/puppet-logrotate/",
+  :ref => 'origin/master'
```
bodepd: for a module like this, I think it makes sense to pin it to a tag or revision so we don't have to worry about it changing out from under us and potentially causing issues. I know that we are not always doing this in our Puppetfile.

manifests/jiocloud/logrotate.pp
```diff
@@ -0,0 +1,18 @@
+define rjil::jiocloud::logrotate(
```
bodepd: why both creating this wrapper class? Why not just call logtotate::rule directly? I don't mind having a wrapper class, I just want to make sure we understand our motivations and don't wind up adding extra code that we don't need.

Puppetfile
anshprat: Yes, I remember you asking for using the tag initially. The reason for using master presently is that there is a patch which is in upstream master but not yet tagged.  https://github.com/JioCloud/puppet-rjil/pull/574#issuecomment-97707767

manifests/jiocloud/logrotate.pp
anshprat: https://github.com/JioCloud/puppet-rjil/pull/574#discussion_r29234156  

I am infact trying to make a reusable snippet (see manifests/jiocloud/logrotate.pp in this commit) (Since most of the logrotate configs are gonna be similar). Else will end up making same calls to ::rule with so many confs.

Puppetfile
anshprat: About rebase - yes, I would try for that next time. My previous attempts at rebase in same branch have not been so great. Would try that again before finalising this.

bodepd: you can just specify a revision instead of a tag (they are also immutable), there are already examples of modules where we need unreleased features and are doing this.

bodepd: for something this simple, the easiest workflow is probably:

````
git log # to find the commit before where you start
git reset <that_commit>
git add,git commit # recommit your code as a new commit
git push --force # force push that commit to this branch
````

the other option is to do a interactive rebase using `git rebase -i` and squash all the commits to a single commit

I would note, that in some cases, mainly where a commit series is complicated and review has started, we prefer not to have rebases to make review easier, but I think in this case it make sense.

File: manifests/jiocloud/logrotate.pp (hunk 2 above)
Comment (bodepd) [inline manifests/jiocloud/logrotate.pp]: I can see there is value is created a wrapper to set defaults. It is also possible to do this with a resource default (although I am not a big fan of a global resource default b/c it makes it hard to apply classes individually)

**Comment by anshprat** [comment]
please test this

**Comment by sorenh** [comment]
You changed the mode of keystone.pp to 775. Please change it back.

**Comment by anshprat** [comment]
please test this

**Comment by anshprat** [comment]
please test this

**Comment by anshprat** [comment]
please test this

**Comment by anshprat** [comment]
please test this

**Comment by anshprat** [comment]
please test this

File: manifests/glance.pp [hunk 3]
```diff
@@ -70,6 +70,16 @@
     headers         => $headers,
   }
 
+  rjil::jiocloud::logrotate { 'glance-api':
```
bodepd: why are you wrapping some of the log rotate config in classes, and putting some straight into the service class? It seems a little inconsistent.

manifests/glance.pp
```diff
@@ -70,6 +70,16 @@
...
+    service => 'glance-api',
+    logfile => '/var/log/glance/api.log'
+  }
+
+  rjil::jiocloud::logrotate { 'glance-registry':
+    service => 'glance-registry',
```
bodepd: in the rjil::jiocloud::logrotate defined resource type, it looks like you can save a lot of code by defaulting service to $name

manifests/glance.pp
anshprat: Yes, this was before our session about resource duplicate handling and my failures to get an array working with a define. Will see if can get it all wrapped up within the define itself.

manifests/glance.pp
anshprat: yes, again one of the things that I realised only after cinder logrotate was done. will improve on it today.

manifests/glance.pp
anshprat: So I now have another problem:


$logsarray = ['test1',
              'teste2',
              'test3',
             ]
rjil::jiocloud::logrotate { $logsarray:
    logfile => "/tmp/$name"
}

This creates correct service as $name but the logfile as /tmp/main for all the logs. Any ideas how to fix this? passing $logsarray to logfile causes string concatnation of the elements.

bodepd: you are calling $name from the wrong scope.

anshprat: ah, good point. But then which variable can I use out here when calling the define with array?

bodepd: that is one thing that is really hard to do in Puppet pre 4 (without iteration). You can use $name inside the define, but not outside. You could pass in the directory if you want to assume that the filename defaults to $name.log

anshprat: yeah, that sounds good, thanks.

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

manifests/contrail/server.pp
```diff
@@ -19,4 +19,19 @@
   }
 
   include ::contrail
+
+  $contrail_logs = [  'contrail-analytic-api',
```
bodepd: where did you get this list of services from? When I look in /var/log/contrail, I see

````
-rw-r--r--  1 contrail contrail        0 May 11 18:08 vnc_openstack.err.1
-rw-r--r--  1 contrail adm             0 May 11 18:08 ifmap-server-console.log
-rw-r--r--  1 contrail contrail     4991 May 11 18:10 vnc_openstack.tmp
-rw-r--r--  1 contrail contrail     4992 May 11 18:10 vnc_openstack.err
-rw-r--r--  1 contrail contrail 10485938 May 11 19:12 contrail-control.log.5
-rw-r--r--  1 contrail contrail 10485801 May 11 20:16 contrail-control.log.4
-rw-r--r--  1 contrail contrail 10486070 May 11 21:19 contrail-control.log.3
-rw-r--r--  1 contrail contrail 10486020 May 11 22:22 contrail-control.log.2
-rw-r--r--  1 contrail contrail  1048836 May 11 22:43 dns.log.10
-rw-r--r--  1 contrail contrail  1048868 May 11 22:49 dns.log.9
-rw-r--r--  1 contrail contrail  1048692 May 11 22:56 dns.log.8
-rw-r--r--  1 contrail contrail  1048848 May 11 23:02 dns.log.7
-rw-r--r--  1 contrail contrail  1048886 May 11 23:08 dns.log.6
-rw-r--r--  1 contrail contrail  1048826 May 11 23:15 dns.log.5
-rw-r--r--  1 contrail contrail  1048865 May 11 23:21 dns.log.4
-rw-r--r--  1 contrail contrail 10485967 May 11 23:25 contrail-control.log.1
-rw-r--r--  1 contrail contrail  1048844 May 11 23:28 dns.log.3
-rw-r--r--  1 contrail contrail  1048873 May 11 23:34 dns.log.2
-rw-r--r--  1 contrail contrail  1048882 May 11 23:40 dns.log.1
drwxr-x---  2 contrail adm          4096 May 11 23:40 .
-rw-r--r--  1 contrail contrail  4425415 May 11 23:44 api.log
-rw-r--r--  1 contrail contrail  3909645 May 11 23:44 api-0-zk.log
-rw-r--r--  1 contrail contrail   377955 May 11 23:44 schema-zk.log
-rw-r--r--  1 contrail contrail  3060852 May 11 23:44 schema.log
-rw-r--r--  1 contrail contrail  1515973 May 11 23:44 discovery.log
-rw-r--r--  1 contrail adm       4697308 May 11 23:44 ifmap-server.log
-rw-r--r--  1 contrail contrail   605399 May 11 23:44 dns.log
-rw-r--r--  1 contrail contrail  3250388 May 11 23:44 contrail-control.log
````

File: manifests/glance.pp [hunk 6]
```diff
@@ -70,6 +70,13 @@
     headers         => $headers,
   }
 
+  $glance_logs = ['glance-api',
```
bodepd: this does not look correct, you will probably have to explicitly specify the lognames b/c they do not follow convention:

````
:/var/log/glance# ls
api.log  registry.log
````

File: manifests/neutron.pp [hunk 7]
```diff
@@ -30,6 +30,10 @@
   include ::neutron::server
   include ::neutron::quota
 
+  rjil::jiocloud::logrotate { 'neutron-server':
```
bodepd: this looks wrong:

````
ls /var/log/neutron/
neutron-netns-cleanup.log  server.log
````

File: manifests/contrail/server.pp (hunk 5 above)
Comment (anshprat) [inline manifests/contrail/server.pp]: huh! This is weird. I used the existing logrotate configs as source of truth. I did look up and compare a sample of nodes and services but not all I guess. Will cross check.

**Comment by bodepd** [comment]
one other thing to add. You might want to also look at rotating the upstart logs, or least have a look in /var/log/upstart to see if they are getting out of hand.

**Comment by anshprat** [comment]
yes, I had checked upstart, they are being rotated correctly by a *.log upstart logrotate config (I guess from the package). I am first targeting the openstack related logs, will then puppetize the rest of system configs.

File: manifests/contrail/server.pp (hunk 5 above)
Comment (anshprat) [inline manifests/contrail/server.pp]: Most of contrail logs are managed by the default log4cxx module via the respective package. The defaults are really low (few M size and count 10) - which gives us only a couple of hours of log - for example for contrail-dns.

To fix this correctly, as per my discussion in opencontrail irc, we will need to create respective logger property files and point the respective confs to these properties and then use logrotate to manage the files.

http://paste.openstack.org/show/220587/

This above task is outside the scope of logrotate and hence will do it under another PR.

File: manifests/neutron.pp (hunk 7 above)
Comment (anshprat) [inline manifests/neutron.pp]: fixed the server.log. the netns-cleanup.log looks mostly like part of the issue discussed above, so will need to handle it separately.

**Comment by anshprat** [comment]
please test this

**Comment by anshprat** [comment]
please test this

**Comment by anshprat** [comment]
please test this

**Comment by anshprat** [comment]
please test this

File: manifests/jiocloud/logrotate.pp [hunk 8]
```diff
@@ -1,14 +1,24 @@
...
+  $logdir        = "/var/log",
+  $logfile       = "notset",
+  $service       = $name,
   $rotate_every  = 'daily',
   $rotate        = 60,
   $compress      = true,
   $delaycompress = true,
   $ifempty       = false,
 ) {
+  if ($logfile == "notset"){
```
bodepd: I'm not a huge fan of picking a string value to mean not set, it's more in line with conventions to use false or undef
in which case you can just say:

````
if $var {

}
````

bodepd:
resubmitted as #603 with unit tests

PR #595, opened 2015-05-12
Collectd Collection

amar266:
1.Installs Collectd
2.Configure collectd for thresholds

### DIFF
--- a/manifests/system/define_metrics.pp
+++ b/manifests/system/define_metrics.pp
@@ -1,0 +1,17 @@
+define rjil::system::define_metrics(
+  $instance,
+  $persist,
+  $persistok,
+  $disks = [],
+  $disk_percent_warn = undef,
+  $disk_percent_fail = undef,
+  $memory_warn = undef,
+  $cpu_warn = undef,
+) {
+
+  #Generate a file
+  file { "/etc/collectd/conf.d/20-thresholds_$name.conf":
+    content => template("rjil/collectd/thresholds-$name.conf.erb"),
+    notify  => Service['collectd'],
+  }
+}
--- a/manifests/system/metrics.pp
+++ b/manifests/system/metrics.pp
@@ -1,0 +1,58 @@
+#
+# [notification_type] Type used to write notifications. Currnetly only supports
+# log.
+#
+class rjil::system::metrics(
+  $notification_type = 'log',
+  $log_file          = '/usr/lib/jiocloud/metrics/collectd_notifications.log',
+) {
+
+  # remove all default plugins so that we can fully customize
+  class { '::collectd':
+    purge        => true,
+    recurse      => true,
+    purge_config => true,
+  }
+
+  # write data to csv files
+  file { ['/var/lib/metrics', '/var/lib/metrics/collectd', '/var/lib/metrics/collectd/csv', '/usr/lib/jiocloud/metrics']:
+    ensure => directory,
+  }
+
+  class { 'collectd::plugin::csv':
+    datadir    => '/var/lib/metrics/collectd/csv',
+    storerates => false,
+  }
+
+  file { '/etc/collectd/conf.d/20-notifications.conf':
+    content => template("rjil/collectd/${notification_type}_notifications.conf.erb"),
+    notify  => Service['collectd'],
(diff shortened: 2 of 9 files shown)


manifests/system/define_metrics.pp
```diff
@@ -0,0 +1,17 @@
...
+  $disks = ['root'],
+  $disk_percent_warn = '10',
+  $disk_percent_fail = '5',
+  $memory_warn = '100000000',
+  $cpu_warn = '10',
+) {
+
+  #Generate a file
+  file { "/etc/collectd/conf.d/20-thresholds_$name.conf":
+      content => template("rjil/collectd/thresholds-$name.conf.erb"),
```
bodepd: I'm really surprised this is passing lint tests This should be 2-space.

manifests/system/metrics.pp
```diff
@@ -0,0 +1,67 @@
...
+
+  include collectd::plugin::memory
+  #include collectd::plugin::thresold
+
+  class { 'collectd::plugin::df':
+    valuespercentage => true
+  }
+
+  # register a consul service that we can use to send out alerts to
+  file { '/usr/lib/jiocloud/metrics/check_thresholds.py':
```
bodepd: this file does not exist in this patch.

manifests/system/define_metrics.pp
```diff
@@ -0,0 +1,17 @@
+define rjil::system::define_metrics(
+  $instance,
```
bodepd: if this is a generic defined resource type for defining metrics, then why does it have the hard-coded values for the types of things that can be passed to it?

bodepd:
some basic unit tests would have caught the notes issues.

manifests/system/define_metrics.pp
amar266: Changes fixed values to undef

manifests/system/metrics.pp
```diff
@@ -0,0 +1,62 @@
+#
+# [notification_type] Type used to write notifications. Currnetly only supports
+# log.
+#
+class rjil::system::metrics(
+  $notification_type = 'log',
+  $log_file          = '/usr/lib/jiocloud/metrics/collectd_notifications.log',
+) {
+
+  user { 'alert-user':
```
bodepd: Is this user even used for anything anymore?

manifests/system/define_metrics.pp
bodepd: I would prefer if this defined resource type didn't even exist. I think it's adds an abstraction layer without much value. If it was a wrapper in front of the same template it might be different, but the fact that it has N sets of parameters that each map to a specific implementation of the configurable template indicate bad design. Just call the template resource directly from the calling class and use a resource default to set common parameters.

amar266: @bodepd  in this defined resource we will be adding consul service codes too
  file { "/etc/collectd/conf.d/20-thresholds_$name.conf":
      content => template("rjil/collectd/thresholds-$name.conf.erb"),
      notify => Service['collectd'],
  }
  rjil::jiocloud::consul::service { "metric_thresholds_$name":
    interval     => '120s',
    tags          => ['metrics'],
    check_command => "/usr/lib/jiocloud/metrics/check_thresholds_$name.sh",
  }
https://github.com/amar266/puppet-rjil/blob/cunsul_collectd/manifests/system/define_metrics.pp

manifests/system/metrics.pp
amar266: I will remove it.

manifests/system/define_metrics.pp
bodepd: fair enough. for now this is fine, but it might be better to accept a generic hash in the future.

bodepd:
code looks fine, but this is blocked until the gate testing environment is operational again. @mayankkapoor please let us know when we can resume testing code.

bodepd:
please test this

bodepd:
please test this

bodepd:
please test this

bodepd:
please test this

bodepd:
rebased, modified, and merged as #610

PR #596, opened 2015-05-12
Consul service configuration on collectd checks

amar266:
1.Creates consul service 
2.create scripts for consul service


bodepd:
rebased as #614

PR #601, opened 2015-05-14
Provide an option to disable non-admin image uploads

varunarya10:
please test this

Do NOT merge

### DIFF
--- a/manifests/glance.pp
+++ b/manifests/glance.pp
@@ -19,8 +19,8 @@ class rjil::glance (
   $ssl                            = false,
   $rewrites                       = undef,
   $headers                        = undef,
-  $allow_add_img_admin_only       = true,
-  $allow_del_img_admin_only       = true,
+  $allow_upload_img_admin_only    = true,
+  $allow_copy_img_admin_only      = true,
   $allow_publicize_img_admin_only = true
 ) {
 
--- a/manifests/glance.pp
+++ b/manifests/glance.pp
@@ -20,7 +20,6 @@ class rjil::glance (
   $rewrites                       = undef,
   $headers                        = undef,
   $allow_upload_img_admin_only    = true,
-  $allow_copy_img_admin_only      = true,
   $allow_publicize_img_admin_only = true
 ) {
 
--- a/manifests/glance.pp
+++ b/manifests/glance.pp
@@ -1,24 +1,27 @@
 ## Class: rjil::openstack::glance
 class rjil::glance (
-  $ceph_mon_key            = undef,
-  $backend                 = 'file',
-  $rbd_user                = 'glance',
-  $ceph_keyring_file_owner = 'glance',
-  $ceph_keyring_path       = '/etc/ceph/keyring.ceph.client.glance',
-  $ceph_keyring_cap        = 'mon "allow r" osd "allow class-read object_prefix rbd_children, allow rwx pool=images"',
-  $admin_email             = 'root@localhost',
+  $ceph_mon_key                   = undef,
+  $backend                        = 'file',
+  $rbd_user                       = 'glance',
+  $ceph_keyring_file_owner        = 'glance',
+  $ceph_keyring_path              = '/etc/ceph/keyring.ceph.client.glance',
+  $ceph_keyring_cap               = 'mon "allow r" osd "allow class-read object_prefix rbd_children, allow rwx pool=images"',
+  $admin_email                    = 'root@localhost',
 
-  $server_name             = 'localhost',
-  $api_localbind_host      = '127.0.0.1',
-  $api_localbind_port      = '19292',
-  $api_public_port         = '9292',
-  $registry_localbind_host = '127.0.0.1',
-  $registry_localbind_port = '19191',
-  $registry_public_address = '127.0.0.1',
-  $registry_public_port    = '9191',
-  $ssl                     = false,
-  $rewrites                = undef,
-  $headers                 = undef,
+  $server_name                    = 'localhost',
+  $api_localbind_host             = '127.0.0.1',
+  $api_localbind_port             = '19292',
+  $api_public_port                = '9292',
+  $registry_localbind_host        = '127.0.0.1',
+  $registry_localbind_port        = '19191',
+  $registry_public_address        = '127.0.0.1',
+  $registry_public_port           = '9191',
+  $ssl                            = false,
+  $rewrites                       = undef,
+  $headers                        = undef,
+  $allow_add_img_admin_only       = true,
+  $allow_del_img_admin_only       = true,
+  $allow_publicize_img_admin_only = true
 ) {
 
   ## Add tests for glance api and registry
@@ -129,6 +132,11 @@ class rjil::glance (
   rjil::jiocloud::consul::service { 'glance-registry':
     tags          => ['real'],
     port          => $::glance::registry::bind_port,
+  }
+
+  file { "/etc/glance/policy.json":
+    ensure  => file,
+    content => template('rjil/glance_policy.erb'),
   }
 
 }
--- a/manifests/glance.pp
+++ b/manifests/glance.pp
@@ -19,7 +19,7 @@ class rjil::glance (
   $ssl                            = false,
   $rewrites                       = undef,
   $headers                        = undef,
-  $allow_upload_img_admin_only    = true,
+  $allow_upload_img_admin_only    = true
 ) {
 
   ## Add tests for glance api and registry
--- a/manifests/glance.pp
+++ b/manifests/glance.pp
@@ -136,7 +136,9 @@ class rjil::glance (
 
   file { "/etc/glance/policy.json":
     ensure  => file,
+    owner   => 'root',
+    mode    => '0644',
     content => template('rjil/glance_policy.erb'),
+    notify  => Service['glance-api'],
   }
-
 }
--- a/manifests/glance.pp
+++ b/manifests/glance.pp
@@ -20,7 +20,6 @@ class rjil::glance (
   $rewrites                       = undef,
   $headers                        = undef,
   $allow_upload_img_admin_only    = true,
-  $allow_publicize_img_admin_only = true
 ) {
 
   ## Add tests for glance api and registry
--- a/templates/glance_policy.erb
+++ b/templates/glance_policy.erb
@@ -1,0 +1,22 @@
+{
+  "context_is_admin":  "role:admin",
+  "default": "",
+<% if @allow_add_img_admin_only == True %>
+  "add_image": "role:admin",
+<% else %>
+  "add_image": "",
+<% end %>
+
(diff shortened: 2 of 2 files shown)


templates/glance_policy.erb
```diff
@@ -0,0 +1,22 @@
+{
+  "context_is_admin":  "role:admin",
+  "default": "",
+<% if @allow_add_img_admin_only == True %>
```
bodepd: templates are ruby under the covers, so use true, not True

varunarya10:
please test this

anshprat:
please test this

varunarya10:
please test this

manifests/glance.pp
```diff
@@ -131,4 +134,9 @@
     port          => $::glance::registry::bind_port,
   }
 
+  file { "/etc/glance/policy.json":
```
alokjani: Glance API service needs a restart/reload for any new policies to be in effect. 
Might require some notification chaining here.

varunarya10:
please test this

bodepd:
the last build failed for two few reasons:
- known issue with vrouters. @varunarya10 just be aware that about 5% of all VM builds lose 1-2 nodes b/c they never get a network route assigned due to an issue with the contrail vrouter, there is effort being done to resolve this issue. However, this only failed oc1 which isn't required to pass all validation tests.
- There is a known issue with a lack of retries for db_sync. It was captured as ticket #85 . It's just a coincidence that it happened to be glance that did not initialize it's database. It's strange that you saw this issue, b/c it's not something we ever see, it seemed to happen b/c of a temporary failure of the hostname of the database to resolve in consul.

There are two issue for ops to address here.

* ensure that we do retry db migrations in case of failure
* understand why consul DNS temporarily failed to resolve

I can see the following log line right before the db_sync failure:

````
 2814 May 15 13:14:58 ocdb1-testjenkins-puppet-rjil-gate-1243 dnsmasq[24673]: reading /etc/resolv.conf
 2815 May 15 13:14:58 ocdb1-testjenkins-puppet-rjil-gate-1243 dnsmasq[24673]: using nameserver 127.0.0.1#53
````

dnsmasq gets reloaded for some reason in the middle of the run which causes a temporary DNS outage

**Comment by bodepd** [comment]
please test this

**Comment by bodepd** [comment]
test failed b/c of vrouter issue on cp15. I finally disabled that node.

**Comment by bodepd** [comment]
please test this

**Comment by bodepd** [comment]
uggh. over the quota limits.

**Comment by bodepd** [comment]
please test this

**Comment by varunarya10** [comment]
please test this.

**Comment by varunarya10** [comment]
please test this

**Comment by varunarya10** [comment]
please test this

File: manifests/glance.pp (hunk 2 above)
Comment (varunarya10) [inline manifests/glance.pp]: As per this link
http://docs.openstack.org/openstack-ops/content/projects_users.html

The policy engine reads entries from the policy.json file. The actual location of this file might vary from distribution to distribution: for nova, it is typically in /etc/nova/policy.json. You can update entries while the system is running, and you do not have to restart services.

**Comment by hkumarmk** [comment]
I see an issue with restricting creating image on to admins - this can lead to a situation where nobody (as non-admin users) can create snapshot, because snapshots are stored in glance as image.

**Comment by hkumarmk** [comment]
may be it is not the case, but we may have to test it - we have done this before in havana, and we face the same problem at that time.

**Comment by varunarya10** [comment]
please test this

**Comment by varunarya10** [comment]
Not sure why the build is failing again. This pull request addresses following bug
https://bugs.launchpad.net/jio/+bug/1452987

- New policies are added which only allow admin to create, delete or publicize images
- When policy.json changes the service glance-api should be notified and service should restart

Concern:
nova image-create internally use glance image-create which is not working because glance does not allow non admin users. Because of this taking a snapshot will not work.

- Preventing users from creating, deleting and publicizing images is a launch critical requirement. We need to address the snapshot issue in a decoupled way. Tracking it here https://bugs.launchpad.net/jio/+bug/1456252

Please merge this once the build passes.

**Comment by varunarya10** [comment]
FYI
Creating snapshots of block volume (cinder) is still allowed and is working even after this change.

**Comment by sorenh** [comment]
No. Just no.

We're not going to merge a patch to address one launch critical requirement well knowing that we'll break another launch critical requirement. I'd be surprised if the product management team is willing to launch without snapshots. It would certainly be the first I've heard of it.

Disabling this functionality shouldn't be the default. Please make the default false and we can override it as appropriate.

Please make your commit messages (as well as PR title+description) explain what you're trying to achieve rather than how you're doing it. Something like "Provide an option to disable non-admin image uploads".

**Comment by varunarya10** [comment]
please test this

Do not merge yet.

File: manifests/glance.pp (hunk 2 above)
Comment (ajayaa) [inline manifests/glance.pp]: The policy file is read again if it changes, in the latest master. Since we are not using master, we might need to check this on the particular release we are using. Afaik, it's there since long time. So perhaps we don't need line number 142.

Comment (varunarya10) [inline manifests/glance.pp]: The policies do not take effect in the current release until glance-api is restarted and hence line 142. The link above does not hold true for the release that we are using.

**Comment by varunarya10** [comment]
please test this

**Comment by varunarya10** [comment]
please test this

**Comment by bodepd** [comment]
looking at your last two builds, 1268 failed due to a non-recoverable failures from userdata, I opened an issue to track it (and submitted a patch to make those kinds of failures recoverable: https://bugs.launchpad.net/jio/+bug/1457169 ). 1267 actually passed if you want to use it for continued dev. I don't understand why jenkins failed.

**Comment by varunarya10** [comment]
please test this.

Based on chat with @anshprat  I am launching a new build.

**Comment by anshprat** [comment]
please test this

**Comment by varunarya10** [comment]
This build is finally passing (after struggling through many dev-ops activities). Thanks to @anshprat.

Based on all the comments from Soren and talking to product management I have altered the policies to allow only admin to upload/copy/publicize image. This would prevent users from creating their own images (from file as well as urls) but at the same time allow the users to create snapshots of their VM. I have tested in my gate environment by creating VM snapshots and launch an instance from the snapshot. 

I have also removed the delete admin only policy as the user should be allowed to delete their own images (which are created as a result of snapshots)

This pull request should now be complete and will address the following:
https://bugs.launchpad.net/jio/+bug/1452987
https://bugs.launchpad.net/jio/+bug/1446601
https://bugs.launchpad.net/jio/+bug/1446606
https://bugs.launchpad.net/jio/+bug/1456252

Please review and merge.

File: templates/glance_policy.erb [hunk 3]
```diff
@@ -0,0 +1,20 @@
+{
+  "context_is_admin":  "role:admin",
+  "default": "",
+<% if @allow_upload_img_admin_only == true %>
```
hkumarmk: you should use <%- (add - after <% ) to avoid adding blank lines in the resulting file. This is true for all conditional expressions (like else, end and conditional lines)

manifests/glance.pp
```diff
@@ -1,24 +1,27 @@
...
+  $api_localbind_port             = '19292',
+  $api_public_port                = '9292',
+  $registry_localbind_host        = '127.0.0.1',
+  $registry_localbind_port        = '19191',
+  $registry_public_address        = '127.0.0.1',
+  $registry_public_port           = '9191',
+  $ssl                            = false,
+  $rewrites                       = undef,
+  $headers                        = undef,
+  $allow_upload_img_admin_only    = true,
```
hkumarmk: Are we going to have a usecase to enable/disable upload_image, copy_from and publicize_image separately? I dont think, so we could just have one param may be something like non_admin_image_upload = true or false and based on that we could just create the file.

manifests/glance.pp
hkumarmk: If we are just have one class param instead of separate param for upload/copy_from/publisize, we just need this to be a static file (not template) and you can add the condition here based on which the file can be created.

hkumarmk:
please test this

hkumarmk:
please test this

bodepd:
functionally is looks fine, I'd just like to get comments from @hkumarmk and @sorenh about the implications for other user functionality that might be required. I think we can do the requested code cleanup as a separate commit.

varunarya10:
please test this

varunarya10:
@bodepd: Thanks for looking into it here is a summary to address all the commants.

Based on comments from @sorenh @hkumarmk I have made the following changes:
1. Clubbed image-upload and image-copy together and they can be controlled by one flag. Default action is to allow only admin to perform this action.
2. Only admin should publicize the image. This action should not be allowed for the customers.
3. Customers will be able to create snapshots of their VMs
4. Changed <% to <%- to avoid blank lines.

Since public cloud is not functional right now. My builds are not running.

Please review and merge.

ajayaa:
Should we not add some functional tests to test this functionality? This functionality depends on the Keystone role name and it could change to suit our needs in future. Also the role name "admin" could be fetched from hiera or someplace common so that if we want to change it in future then we would need to change it in one place only.

I am not sure what would be a good place to add the tests. Maybe tempest or someplace where there would be jio deployment specific tests.

varunarya10:
@ajayaa  This is initial configuration for production in terms of policy. If we want to change the role name from admin or something similar then we would already break many things as there are already similar configuration that exists in this file (i.e. "role:admin" without these changes). I have tested these changes in devstack and in the gate environment. If there is a need to add tempest test then we can add it in a separate pull request. For now it is one of the launch requirement.

I have made all the changes that were advised/suggested. I just need to launch one final build  to make sure everything works. Otherwise it is ready to be reviewed and merged.

templates/glance_policy.erb
```diff
@@ -0,0 +1,13 @@
+{
+  "context_is_admin": "role:admin",
+  "default": "",
+<%- if @allow_upload_img_admin_only == true %>
```
bodepd: this doesn't look right. the '-' needs to go before the closing text ( -%> ), I'd expect what you have here to be a syntax error.

also for the case of true, you can just check

````
<% if @allow_upload_img_admin_only -%>
````

varunarya10:
please test this.

varunarya10:
@hkumarmk @bodepd 
I created build 1307 http://jiocloud.rustedhalo.com:8080/job/puppet-rjil-gate/1307/ which passed.
The policy.json generated is valid and without unnecessary newlines. There is only one flag now.
I have verified in the gate environment:
1. JSON is valid
2. I can not upload images from files or urls as a non admin role
3. I can create VM snapshots from my VMs. (nova image-create works)

Please review and merge.

varunarya10:
Hi. This request is still pending to be merged. Please review and merge.

sorenh:
Just to clarify, are the defaults values the desired values for production or do we need to override anything for production?

sorenh:
I'm sorry this has taken so long to get merged. I'm ready to merge it once my question from a few minutes ago is clarified.

varunarya10:
@sorenh Yes the defaults are the desired values. We do not need to override anything for production.

PR #602, opened 2015-05-14
contrail logs

anshprat:
pointing to my repo for testing

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -32,8 +32,7 @@ fixtures:
       repo: https://github.com/saz/puppet-ssh
       ref: v2.3.6
     contrail:
-      repo: https://github.com/anshprat/jiocloud-contrail
-      ref: logs_1
+      repo: https://github.com/jiocloud/jiocloud-contrail
     sudo:
       repo: https://github.com/saz/puppet-sudo
       ref: 62b93da
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -32,7 +32,8 @@ fixtures:
       repo: https://github.com/saz/puppet-ssh
       ref: v2.3.6
     contrail:
-      repo: https://github.com/jiocloud/jiocloud-contrail
+      repo: https://github.com/anshprat/jiocloud-contrail
+      ref: logs_1
     sudo:
       repo: https://github.com/saz/puppet-sudo
       ref: 62b93da
--- a/files/contrail-consolidate-logs.sh
+++ b/files/contrail-consolidate-logs.sh
@@ -1,9 +1,20 @@
 #!/bin/bash
+##
+# The contrail codebase we use does not support log configuration for most
+# logs. The default is rotated based upon size. The non-configurable default
+# size is too less (few MBs) and hence with the default retention of 10 logs,
+# we end up with only few hours of logs for these. Hence this script collates
+# these smaller files into a single daily logfile which can then be rotated
+# via logrotate. The upstream has added config options for some modules but not
+# all are still covered. Hence till upstream is completely patched, this will
+# still be useful
+##
 
 contrail_logs=('contrail-api'
 		'contrail-schema'
 		'contrail-vrouter'
 		'contrail-discovery'
+		'contrail-collector'
 		'vnc_openstack.err'
 		'svc-monitor.err'
 		'schema.err'
--- a/files/contrail-consolidate-logs.sh
+++ b/files/contrail-consolidate-logs.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+contrail_logs=('contrail-api'
+		'contrail-schema'
+		'contrail-vrouter'
+		'contrail-discovery'
+		'vnc_openstack.err'
+		'svc-monitor.err'
+		'schema.err'
+		'api-0-zk'
+		'schema-zk'
+		)
+
+log_dir='/var/log/contrail/'
+for log in ${contrail_logs[@]} 
+do 
+file="${log_dir}${log}"
+file_log="${file}.log.1"
+file_daily="${file}-daily.log"
+
+if [[ "$log" =~ "err" ]]
+then
+    file_log="${file}.1"
+    file_daily="${log_dir}contrail-${log}-daily.log"
+fi
+ 
+if [[ "$log" =~ "zk" ]]
+then
+    file_daily="${log_dir}contrail-${log}-daily.log"
+fi
+
+if [ -f "${file_log}" ]
+then
+	cat $file_log >> $file_daily
+	rm  $file_log
+fi
+
+done
--- a/manifests/contrail/logrotate/consolidate.pp
+++ b/manifests/contrail/logrotate/consolidate.pp
@@ -1,0 +1,13 @@
+class rjil::contrail::logrotate::consolidate {
+  file { '/usr/local/bin/contrail-consolidate-logs.sh':
+    source => 'puppet:///modules/rjil/contrail-consolidate-logs.sh',
+    mode   => '0755',
+    owner  => 'root',
+    group  => 'root'
+  }
+  cron { 'contrail-consolidate-logs':
+    command => 'run-one /usr/local/bin/contrail-consolidate-logs.sh 2>&1 | logger',
+    user    => 'root',
+    require => Package['run-one'],
+  }
+}
--- a/manifests/contrail/server.pp
+++ b/manifests/contrail/server.pp
@@ -20,18 +20,35 @@ class rjil::contrail::server (
 
   include ::contrail
 
-  $contrail_logs = [  'contrail-analytic-api',
-                      'contrail-collector',
-                      'query-engine',
-                      'api',
-                      'discovery',
-                      'schema',
-                      'svc-monitor',
-                      'contrail-control',
-                      'webserver',
-                      'jobserver',
+  $contrail_logs = ['contrail-api-daily',
+                    'contrail-discovery-daily',
+                    'contrail-schema-daily',
+                    'contrail-svc-monitor-daily',
+                    'contrail-ifmap-server'
   ]
+
   rjil::jiocloud::logrotate { $contrail_logs:
     logdir => '/var/log/contrail'
   }
+  
+  $contrail_logs_copytruncate = ['contrail-control',
+                                'contrail-dns',
+  
+  ]
+  
+  rjil::jiocloud::logrotate { $contrail_logs_copytruncate:
+    logdir       => '/var/log/contrail',
+    copytruncate => true,
+  }
+  
+  include rjil::contrail::logrotate::consolidate
+
+  $contrail_logrotate_delete = ['contrail-config',
+                                'contrail-config-openstack',
+                                'ifmap-server',
+                                ]
+  rjil::jiocloud::logrotate { $contrail_logrotate_delete:
+    ensure => absent
+  }
+  
 }
--- a/manifests/contrail/server.pp
+++ b/manifests/contrail/server.pp
@@ -24,16 +24,23 @@ class rjil::contrail::server (
                     'contrail-discovery-daily',
                     'contrail-schema-daily',
                     'contrail-svc-monitor-daily',
-                    'contrail-ifmap-server'
+                    'contrail-api-0-zk-daily',
+                    'contrail-collector-daily'
   ]
 
   rjil::jiocloud::logrotate { $contrail_logs:
     logdir => '/var/log/contrail'
   }
   
+  ##
+  # The logs which support higher filesize need either a sighup or
+  # a copytruncate to rotate properly (else the process will keep writing to
+  # older logfile. Using copytruncate to minimize any potential issue w/ sighup
+  ##
+
   $contrail_logs_copytruncate = ['contrail-control',
                                 'contrail-dns',
-  
+                                'contrail-ifmap-server',
   ]
   
   rjil::jiocloud::logrotate { $contrail_logs_copytruncate:
@@ -43,8 +50,14 @@ class rjil::contrail::server (
   
   include rjil::contrail::logrotate::consolidate
 
+  ##
+  # Deleting the default config logrotates which conflict with our changes
+  # The default configs have multiple logfiles in a single config which
+  # conflicts with our daily files setup
+  ##
   $contrail_logrotate_delete = ['contrail-config',
                                 'contrail-config-openstack',
+                                'contrail-analytics',
                                 'ifmap-server',
                                 ]
   rjil::jiocloud::logrotate { $contrail_logrotate_delete:
--- a/manifests/jiocloud/logrotate.pp
+++ b/manifests/jiocloud/logrotate.pp
@@ -1,14 +1,17 @@
 define rjil::jiocloud::logrotate(
   $logdir        = "/var/log",
-  $logfile       = "notset",
+  $logfile       = 'undef',
   $service       = $name,
   $rotate_every  = 'daily',
   $rotate        = 60,
   $compress      = true,
   $delaycompress = true,
   $ifempty       = false,
+  $copytruncate  = 'undef',
+  $dateext       = true,
+  $ensure        = 'present',
 ) {
-  if ($logfile == "notset"){
+  if ($logfile == 'undef'){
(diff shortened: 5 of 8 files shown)


anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

.fixtures.yml
```diff
@@ -32,7 +32,8 @@ fixtures:
       repo: https://github.com/saz/puppet-ssh
       ref: v2.3.6
     contrail:
-      repo: https://github.com/jiocloud/jiocloud-contrail
+      repo: https://github.com/anshprat/jiocloud-contrail
+      ref: logs_1
```
anshprat: These are tmp changes for testing, to be reverted before merge after jiocloud-contrail merge

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
@hkumar @soren @bodepd This is ready for review

files/contrail-consolidate-logs.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
hkumarmk: What is the motive of this script to move the errors and other stuffs to daily log?

hiera/data/common.yaml
```diff
@@ -643,6 +643,26 @@ contrail::enable_analytics: false
 
 rjil::contrail::server::enable_analytics: false
 
+log_level: SYS_INFO
```
hkumarmk: All these params are default params in contrail module, so no need to pass them through hiera, let them use class defaults.

files/contrail-consolidate-logs.sh
anshprat: These logs are presently size bound to the tune of few MBs. Hence presently we only have a couple of hours worth of data. So we first aggregate the data to a daily file and let logrotate handle the daily file. Our codebase at this stage does not allow logging config modifications for most of contrail modules. Upstream is now slighty better (patched april end) but not all modules are covered yet.

manifests/contrail/server.pp
```diff
@@ -20,18 +20,35 @@
...
-                      'contrail-collector',
-                      'query-engine',
-                      'api',
-                      'discovery',
-                      'schema',
-                      'svc-monitor',
-                      'contrail-control',
-                      'webserver',
-                      'jobserver',
+  $contrail_logs = ['contrail-api-daily',
```
hkumarmk: I am bit confused about these "daily" logs, why do we need them like this?

manifests/contrail/logrotate/consolidate.pp
```diff
@@ -0,0 +1,13 @@
+class rjil::contrail::logrotate::consolidate {
```
hkumarmk: why do we need to consolidate the logs? can't it be done through log configuration?

manifests/jiocloud/logrotate.pp
```diff
@@ -1,14 +1,17 @@
 define rjil::jiocloud::logrotate(
   $logdir        = "/var/log",
-  $logfile       = "notset",
+  $logfile       = 'undef',
```
hkumarmk: undef without quotes

manifests/jiocloud/logrotate.pp
```diff
@@ -1,14 +1,17 @@
...
   $rotate        = 60,
   $compress      = true,
   $delaycompress = true,
   $ifempty       = false,
+  $copytruncate  = 'undef',
+  $dateext       = true,
+  $ensure        = 'present',
 ) {
-  if ($logfile == "notset"){
+  if ($logfile == 'undef'){
```
hkumarmk: you just need if !$logfile

manifests/jiocloud/logrotate.pp
```diff
@@ -1,14 +1,17 @@
...
       if ($logdir =~ /\/$/) {
```
hkumarmk: This is not required, there is no harm to just add ${logdir}/${name}.log" irrespective of logdir is end with '/' or not

manifests/contrail/server.pp
```diff
@@ -20,18 +20,35 @@
...
+  ]
+  
+  rjil::jiocloud::logrotate { $contrail_logs_copytruncate:
+    logdir       => '/var/log/contrail',
+    copytruncate => true,
+  }
+  
+  include rjil::contrail::logrotate::consolidate
+
+  $contrail_logrotate_delete = ['contrail-config',
```
hkumarmk: What is this? why you need to remove certain logrotate entries?

manifests/contrail/logrotate/consolidate.pp
anshprat: Our codebase at this stage does not allow logging config modifications for most of contrail modules.

hkumarmk:
@anshprat It would be better you add some inline documentation to brief what and why those codes.

files/contrail-consolidate-logs.sh
bodepd: can we not just make the desired adjustments to the logrotate config? This seems really complicated.

anshprat: @bodepd unfortunately logrotate will only work on top of a given logfile. The issue is with the given logfile presently which is only of few MBs here. So if we just keep rotating them, for retaining 60 days of logs, each day will generate upwards of 24 logs :! Or use two logrotate configs - one to create something like the daily log version, and another to rotate the daily log. Hence creating this script to create a standard file which can be just rotated by logrotate.

manifests/jiocloud/logrotate.pp
anshprat: yeah, not required from practical purpose but added this to keep the config file sames. Else depending upon how the param is provided, will end up with configs having double slashes.

manifests/contrail/server.pp
anshprat: These default config files provided by package have multiple log lines in each config file. Hence some of them override the configs we are creating here. So deleting these default config files.

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

files/contrail-consolidate-logs.sh
bodepd: I'll have a close look at this tomorrow. My instinct is that the fact that you need to model things as a procedure using bash means that you aren't thinking about the problem right (or at least in the Puppet way). I feel like there must be a way to model things as the desired state of the configuration. I can spend some time with a build tomorrow to get in depth enough to try to better understand your motivations here and if there might be a better way.

anshprat:
please test this

anshprat:
please test this

bodepd:
@anshprat if you get a build working that you feel good about, please leave it for me, and I'll jump on and take a look when I wake up.

anshprat:
if only I had a cent for every failed build... well, atleast I could ve a free lunch :/

anshprat:
please test this

anshprat:
@bodepd 1321 is all green.

bodepd:
It might be better if I try to catch you online, I don't quite understand what is going on here :(

Let me try to state what I think the issue is that leads to all of this complexity:

contrail uses log4j as the backend for it's logging, and is also using log4j to manage log rotation. This rotation is based on log file size, and for some services is set so low, that a majority of the log information can't be retained in a way that support our requirements.

To resolve this, you are running a script continuously to append the log contents into a daily log file to each service, then specifying log rotation rules on that file.

To offer an alternative that might be a little easier to implement, can we just disable the logrotation that is happening through the contrail config (maybe by setting it to -1 or something) and then rely entirely on logrotation rules? I'm sure that I'm still missing some details about why that is not possible. I'll try to catch up to you during your Mon or Tues morning.

bodepd:
one more thing, I went through all of the config files and resulting log files by hand, and lots of the log files indicated contrail errors. I cross checked the log files with prod and did not see the same log entries there. I can also cross check against a recent build to see if the errors exist there as well. If these changes are causing contrail to be in a failed state, then our validation scripts are not robust enough in this case to detect it :(

failures in the dns logs:

````
sh: Send: FAILED: 1432926490533535 IFMapStateMachine [SYS_DEBUG]: IFMapSmEventMessage: Processing ifsm::EvConnectionCleaned in state SsrcStart controller/src/ifmap/client/ifmap_state_machine.cc 944
2015-05-29 Fri 19:08:10:533.608 UTC  ct1-testjenkins-puppet-rjil-gate-1321 [Thread 140132601620224, Pid 26806]: Sandesh: Send: FAILED: 1432926490533596 IFMapStateMachine [SYS_DEBUG]: IFMapSmTransitionMessage: SsrcStart ===> ConnectTimerWait controller/src/ifmap/client/ifmap_state_machine.cc 954
2015-05-29 Fri 19:08:10:533.653 UTC  ct1-testjenkins-puppet-rjil-gate-1321 [Thread 140132601620224, Pid 26806]: Sandesh: Send: FAILED: 1432926490533640 IFMapStateMachine [SYS_DEBUG]: IFMapSmStartTimerMessage: 1 second connect timer started. controller/src/ifmap/client/ifmap_state_machine.cc 707
````

/var/log/contrail/contrail-discovery.log
````
05/29/2015 06:54:58 PM [ct1-testjenkins-puppet-rjil-gate-1321:DiscoveryService:Config:0]: No Connection: __default__ [SYS_INFO]: discServiceLog: subscribe: service type=Collector, client=VRouterAgent:cp2-testjenkins-puppet-rjil-gate-1321:VRouterAgent, ttl=32, asked=2 pubs=0/0, subs=0
````

/var/log/contrail/contrail-schema.log

````
05/29/2015 11:01:27 AM [ct1-testjenkins-puppet-rjil-gate-1321:Schema:Config:0]: Sandesh Client: Event[EvCollectorUnknown] => State[Connect] -> State[Disconnect]
05/29/2015 11:02:22 AM [ct1-testjenkins-puppet-rjil-gate-1321:Schema:Config:0]: Processing event[EvSandeshUVESend] in state[Disconnect]
````
control log also has lots of failures:
````
desh: Send: FAILED: 1432927174918091 IFMapStateMachine [SYS_WARN]: IFMapPeerConnError: wrong version number SsrcConnect SsrcSslHandshake ifsm::EvConnectSuccess controller/src/ifmap/client/ifmap_state_machine.cc 916
2015-05-29 Fri 19:19:34:918.154 UTC  ct1-testjenkins-puppet-rjil-gate-1321 [Thread 140144828499904, Pid 26953]: Sandesh: Send: FAILED: 1432927174918146 IFMapStateMachine [SYS_DEBUG]: IFMapSmEventMessage: Enqueuing ifsm::EvHandshakeFailed in state SsrcSslHandshake controller/src/ifmap/client/ifmap_state_machine.cc 937
2015-05-29 Fri 19:19:34:918.222 UTC  ct1-testjenkins-puppet-rjil-gate-1321 [Thread 140144660563712, Pid 26953]: Sandesh: Send: FAILED: 1432927174918210 IFMapStateMachine [SYS_DEBUG]: IFMapSmEventMessage: Processing ifsm::EvHandshakeFailed in state SsrcSslHandshake controller/src/ifmap/client/ifmap_state_machine.cc 944
2015-05-29 Fri 19:19:34:918.258 UTC  ct1-testjenkins-puppet-rjil-gate-1321 [Thread 140144660563712, Pid 26953]: Sandesh: Send: FAILED: 1432927174918250 IFMapStateMachine [SYS_DEBUG]: IFMapSmCancelTimerMessage: Cancelling Response timer. controller/src/ifmap/client/ifmap_state_machine.cc 743
2015-05-29 Fri 19:19:34:918.290 UTC  ct1-testjenkins-puppet-rjil-gate-1321 [Thread 140144660563712, Pid 26953]: Sandesh: Send: FAILED: 1432927174918282 IFMapStateMachine [SYS_DEBUG]: IFMapSmTransitionMessage: SsrcSslHandshake ===> SsrcStart controller/src/ifmap/client/ifmap_state_machine.cc 954
2015-05-29 Fri 19:19:34:918.489 UTC  ct1-testjenkins-puppet-rjil-gate-1321 [Thread 140144828499904, Pid 26953]: Sandesh: Send: FAILED: 1432927174918477 IFMapPeer [SYS_DEBUG]: IFMapServerConnection: Retrying connection to Ifmap-server.  controller/src/ifmap/client/ifmap_channel.cc 204
````

ifmap server:

````
2015-05-29 19:23:50,925 [pool-6-thread-97] ERROR  - ChannelThread: Setting channel 10.0.0.98:47203:20507 into state 'BROKEN'
2015-05-29 19:23:53,921 [pool-6-thread-97] ERROR  - ChannelThread: Receiving request failed
2015-05-29 19:23:53,922 [pool-6-thread-97] ERROR  - ChannelThread: SSLHandshakeException: Client doesn't know about our certificate (?)
2015-05-29 19:23:53,922 [pool-6-thread-97] ERROR  - ChannelThread: Setting channel 10.0.0.98:47218:20508 into state 'BROKEN'
2015-05-29 19:23:53,933 [pool-6-thread-97] ERROR  - ChannelThread: Receiving request failed
2015-05-29 19:23:53,933 [pool-6-thread-97] ERROR  - ChannelThread: SSLHandshakeException: Client doesn't know about our certificate (?)
2015-05-29 19:23:53,933 [pool-6-thread-97] ERROR  - ChannelThread: Setting channel 10.0.0.98:47219:20509 into state 'BROKEN'
````

**Comment by anshprat** [comment]
@bodepd lets sync up on this today?

**Comment by anshprat** [comment]
please test this

File: files/contrail-consolidate-logs.sh [hunk 10]
```diff
@@ -0,0 +1,49 @@
...
+		'schema.err'
+		'api-0-zk'
+		'schema-zk'
+		)
+
+log_dir='/var/log/contrail/'
+for log in ${contrail_logs[@]} 
+do 
+file="${log_dir}${log}"
+file_log="${file}.log.1"
```
bodepd: I just logged in and had a look, this is the piece of data that I was missing before, you are checking the rotated out files. I feel like there is some risk that we might lost log contents here, I'm going to leave a build alive for a day and see what the results look like.

All in all, it doesn't seem as scary in terms of race-conditions as I thought. I'm going to sit down and think about possible race conditions, but for the most part, I think it's okay.

bodepd:
I did some testing and found two issues:

1. the files are saved as root and not as the contrail user (I'm not sure if this matters, but it makes the inconsistent with the rest of the logs)
2. in any case where it adds a number higher than 1, those files are not appended (ie: .2, .3). I know given how often the script runs, this should not happen, but I think we still need logic to protect against this case.

anshprat:
1) - Agreed about consistency. Though I can fix up the ownership as part of this PR, I think fixing the  ownership as part of fixing all (contrail + other) logs to be non-sudo readable would be consistent and easier.
2) - I ve knowingly ignored .2+ since they will come into picture only when .1 is lying around. When this script is first pushed there will be some of these older logs. Adding them to the consolidation script will add complexity and will add a lot of cases to be handled. 

For example - lets say, someone decompresses one of the older pre-script log.5.gz files in the same location. This newly uncompressed log.5 should not be added  to the daily.log. Again, I can add check for time etc, but this adds complexity. 

Another thing to be take care of would be to ensure that the logs are appended to daily in right chronological order.

There is one usecase where it could be useful - this custom script somehow did not run correctly/was paused manually for long enough for .2 to be created.

Logrotate script would generally handle only the specified log file. For the purpose of this script, the specified log file is log.1.

So we have to take a call whether to handle all such situations. Its doable, but I am not sure its worth the various issues that can crop up due to this or worth the complexity it may add. Thoughts?

bodepd:
let's merge it.

PR #606, opened 2015-05-22
adding script for vrouter restart

anshprat:


anshprat:
please test this

anshprat:
please test this

files/vrouter-test-restart.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+log_file="/var/log/contrail/rjil-vrouter-restart.log"
+test_cmd='curl -s -m 5 -o /dev/null http://localhost:8085/Snh_ItfReq?name='
```
mayankkapoor: Question: Does it make more sense to use the "Snh_DhcpInfo? command instead of the "Snh_ItfReq?name=" as discussed on Slack?

anshprat: @mayankkapoor my understanding from the discussion was that Mayank G preferred this api as this gave the actual data and the actual failure. So a dhcpinfo failure does not impact anything but the snh_itfreq failure impacts our issue.

anshprat:
please test this

manifests/contrail/vrouter.pp
```diff
@@ -40,6 +40,18 @@
...
   }
+
+  file { '/usr/local/bin/vrouter-test-restart.sh':
+    source => 'puppet:///modules/rjil/vrouter-test-restart.sh',
+    mode   => '0755',
+    owner  => 'root',
+    group  => 'root'
+  }
+
+  exec { 'vrouter-test-restart':
```
bodepd: Puppet should not be executing scripts like this, it should configure cron

anshprat: @bodepd I thought so, but then the exit 1 part made me move it to exec. Should I put it in cron and a lock file check not to restart?

bodepd: what do you mean a lock check not to restart?

anshprat: basically if the script exits with error, create some lock file. In the beginning of the script, check for the absence of the lock file and proceed only if its absent. But based on what I am seeing, even in exec- if the script exits - puppet will still restart the script is it?

bodepd: moving it to Puppet has a few issues. Puppet only runs when it thinks it has to and even then, execs that run through Puppet have quite a bit of delay as Puppet runs take a while.

anshprat: alright, will move it to cron.

bodepd: what would the purpose of a lock file be? Puppet always runs the exec, although it can be configured to conditionally run.

anshprat: Based upon my understanding, we want to try the router restart if the api fails. But if the restart also fails, then we want the script to exit and not try another restart. The lock file is to prevent further restart attempts till the lock file is cleared?

bodepd: That is reasonable, but I still wouldn't put this logic in Puppet. 2 places to consider.

1. cron - probably makes the most sense so that we can set the frequency, if you want to create a lock file , that logic can still exist in the bash script run from cron.
2. /usr/lib/jiocloud/tests would also be a good location so that we can keep track of failures through consul validation (ie: if a vrouter is in a failed state, the it should fail validation)

anshprat: 1) Oh, I was talking about putting the logic in the bash script, not the puppet. Sorry for the confusion!
2) ok, will move it to the tests part. Will need to check more on that.

anshprat:
please test this

manifests/contrail/vrouter.pp
anshprat: 1) done
2) not sure about the entire flow of consul checks, hence kept it in cron for the time being.

anshprat:
please test this

anshprat:
please test this 
The builds are failing coz of the problem that this patch is trying to mitigate. cp5 and cp7 are facing this issue presently.

anshprat:
please test this

anshprat:
please test this

manifests/contrail/vrouter.pp
```diff
@@ -40,6 +40,20 @@
...
   }
+
+  file { '/usr/local/bin/contrail-vrouter-restart.sh':
+    source => 'puppet:///modules/rjil/contrail-vrouter-restart.sh',
+    mode   => '0755',
+    owner  => 'root',
+    group  => 'root'
+  }
+
+  cron { 'contrail-vrouter-restart':
```
hkumarmk: why you need to run this through cron? you may just run it as part of consul validation, so any failure will cause consul alert.

files/contrail-vrouter-restart.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+log_file="/var/log/contrail/rjil-vrouter-restart.log"
+test_cmd='curl -s -m 5 -o /dev/null http://localhost:8085/Snh_ItfReq?name='
+lock_file='/var/run/contrail-vrouter-restart.err'
+sleep_time=10
+
+while [ 1 ]
```
hkumarmk: why you need to run this in loop? you may just run it once and let whoever run this script run it again?

manifests/contrail/vrouter.pp
anshprat: Cron was one of the ways @bodepd suggested along with cosul way https://github.com/JioCloud/puppet-rjil/pull/606#discussion_r30916910 . I looked into the consul way a bit but wasnt sure of the entire path, hence went with cron.

files/contrail-vrouter-restart.sh
```diff
@@ -0,0 +1,32 @@
...
+    if [ ! -f $lock_file ]
+    then
+        last_step='Curl'
+        ${test_cmd}
+        rv=$?
+        if [ $rv == 28 ]
+        then
+           echo "[INFO] `date` `hostname` [`date +%s`] Restarting vrouter ">>${log_file}
+           last_step='Agent restart'
+           service contrail-vrouter-agent restart
```
hkumarmk: There is a contrail_check script which does a curl to port 8085 - I see you just add this contrail-vrouter-agent restart logic to it.

So like this - 
   check curl is working, 
     curl to the url is not working 
           restart vrouter and run curl
               Curl not working, just exit 1 so consul alert will be there (unless this is a service check exit 2 will cause all services in that node to be down, so just exit 1)
           curl is working, just exit 0

files/contrail-vrouter-restart.sh
anshprat: Based upon discussion, the idea I got is to ensure that as and when the vrouter goes down, try to restart it, and if the restart fails, exit. In cron way, the least possible time is per min, whereas the requirement here is to poll every 10 sec. Hence keeping it in a while loop. This is to minimize manual intervention unless the restart itself fails.

hkumarmk: It should be a consul check.

manifests/contrail/vrouter.pp
hkumarmk: There is one contrail check validation script already there, so you just need to add restart logic to it (https://github.com/JioCloud/puppet-rjil/blob/master/templates/tests/contrail_vrouter.sh.erb)

you may also make it conditional, so you can choose this script to restart vrouter or not.

hkumarmk: oh no, validation script may not work, it should be a different health check  - because validation is a ttl check

bodepd:
I want to clarify that we are holding off review and merging of this patch until we explore a different root cause as determined by @hkumarmk ?

anshprat:
@bodepd yup, this is on hold for the time being.

anshprat:
ok, this is back on. Even though most of the comments from earlier review are still valid - I would need pointers on how to move it to consul check / example check (i.e, if we dont want to do a cron based solution)  @hkumarmk @bodepd

anshprat:
closing this as will be doing via consul check route instead.

PR #611, opened 2015-05-27
Adding tmpreaper to clean puppet reports

hkumarmk:
One of the reason for disk getting filled is puppet reports, just pushing a
patch to add tmpreaper to purge old puppet reports.

### DIFF
--- a/manifests/system.pp
+++ b/manifests/system.pp
@@ -4,6 +4,7 @@
 class rjil::system(
   $proxies                       = {},
   $dhclient_override_domain_name = undef,
+  $puppet_report_keep_hours      = 24,
 ) {
 
   ##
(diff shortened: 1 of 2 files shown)


manifests/system.pp
```diff
@@ -18,7 +19,21 @@
...
 
-  ensure_packages(['molly-guard'])
+  ensure_packages(['molly-guard','tmpreaper'])
+
+  ##
+  # delete puppet reports which are older than $puppet_report_keep_hours hours
+  # Ideally it should have added in tmpreaper daily cron script, but with the
+  # current tmpreaper daily cron script, it is not possible to provide different
+  # timespecs for different directories.
+  # Create a cron to run tmpreaper once every day
```
bodepd: I really hate to see this happen, the puppet reports have so much useful information about what is happening, but I understand the motivation...

hkumarmk: @bodepd are we going to use those reports older than 24 hours?

bodepd: I would like to be able to consider it at some point. We could probably compress them pretty easily with a ruby script.

anshprat:
@bodepd ping

anshprat:
Unless we have the storage node for logs retention/aggregation, there is no alternate place to store the logs. Merging this for the time being , but we should revisit this when we have a place to store the logs.

bodepd:
totally fine. I'm happy for you to make the call especially b/c you understand the data that we are losing.

PR #621, opened 2015-06-02
Limit number of snapshots to 240 per customer account

swamireddy:
Fix https://bugs.launchpad.net/jio/+bug/1454072

Closes-bug: #1454072

### DIFF
--- a/hiera/data/common.yaml
+++ b/hiera/data/common.yaml
@@ -300,6 +300,7 @@ cinder::volume::rbd::volume_tmp_dir: /tmp
 cinder::volume::rbd::rbd_secret_uuid: "%{hiera('cinder_rbd_secret_uuid')}"
 cinder::glance::glance_api_servers: "%{hiera('api_protocol')}://%{hiera('glance_public_address')}:%{hiera('glance_api_port')}"
 cinder::quota::quota_volumes: 120
+cinder::quota::quota_snapshots: 240
 
 ########### Nova config
 ###################################
--- a/manifests/cinder.pp
+++ b/manifests/cinder.pp
@@ -64,6 +64,7 @@ class rjil::cinder (
   $volume_nofile           = 10240,
   $rewrites                = undef,
   $headers                 = undef,
+  $use_default_quota_class = false,
 ) {
 
   ######################## Service Blockers and Ordering
@@ -109,6 +110,11 @@ class rjil::cinder (
   ##
 
   cinder_config { 'DEFAULT/osapi_volume_listen_port': value => $localbind_port }
+
+  ##
+  # Cinder default quotas read from the config file.
+
+  cinder_config { 'DEFAULT/use_default_quota_class': value => $use_default_quota_class }
 
   ## Configure apache reverse proxy
   apache::vhost { 'cinder':


swamireddy:
please test this

bodepd:
please test this

bodepd:
I had a look at this patch and verified that it does set up the correct quota values for installation, but it is not working for updating (which will be required for production to work correctly). It looks as if updating quota values in cinder.conf and restarting services does not work. @sorenh do you know anything about this? I was just about to dig around in the code and see if I can find some answers.

swamireddy:
@bodepd: Can you please share the error msg to debug the issue?

bodepd:
create a build using one value (in your example, it's set to 240)
update the value in cinder.conf (I tried setting it to 250)
use cinder quota-show to verify that it doens't update.

bodepd:
I left a comment in an opened upstream ticket but haven't gotten a response. I started digging around in the code to try to understand the behavior, but couldn't find the exact place where the values are being persisted.

I'm guessing based on the behavior that I saw that updates aren't honored, and that update is only support through command line tools. This needs to be confirmed though.

bodepd:
let me take another stab at this tomorrow. After talking with Soren, there are a few more things that I can look into.

swamireddy:
please test this

swamireddy:
@bodepd: Any update on this patch?

bodepd:
@swamireddy Why would I have an update, I thought you were working on this? Why have you not tried to add code to perform updates?

swamireddy:
@bodepd: Thanks for support. 
Yes, I have worked on it and fixed the issue in the cinder code (yet submit the patch). With the fix, cinder reads the conf with updated the default values.

bodepd:
While I think that fixing the cinder source is the right long term solution, I would recommend just implementing the update calls as exec resources in puppet for the interim. Once we have that patch in, devops, we can have a discussion about if another approach might be required.

swamireddy:
Iam +1 with cinder code changes as a solution for this issue and its clean and one time fix.

swamireddy:
Before this patch test and merge, the below patch has to be applied at puppet-cinder repo.
https://github.com/JioCloud/puppet-cinder/pull/2

anshprat:
@swamireddy to test this with your changes , you need to point it to your repo temporarily for testing. example - https://github.com/anshprat/puppet-rjil/commit/9aa30ad25c8ee630f5a903f7f0d13cf604dc969c  https://github.com/JioCloud/puppet-rjil/pull/602/commits

swamireddy:
@anshprat:  Here is the repo for testing:

  https://github.com/swamireddy/puppet-cinder/commit/02e5e86791dd958c8afb584f845f85d032733502

hkumarmk:
@swamireddy you may just change Puppetfile and .fixtures.yml in puppet-rjil to point to your branch and test the change. Once it pass, the other change will be merged and then you can revert your changes and can merge this.

swamireddy:
please test this

swamireddy:
@anshprat: Looks like all tests passed.  what is next step?

swamireddy:
Can I get this build VM or Env for testing?

bodepd:
You don't have to patch puppet-cinder with that code, you could also just add that same resource to rjil::cinder.

It's definitely worth submitting a patch upstream to get it accepted, but until it is. It's easier for us to carry the change in puppet-rjl.

I'd also like to see more documentation. Just looking at the code changes, I don't entirely understand what that change does. If it doesn't use the specified quota class, then what does it use?

bodepd:
@anshprat please read the above comment. For things that make sense, I'd rather carry the patch in puppet-rjil so that we don't have to fork the upstream repos. We should, of course, also submit the patches upstream.

swamireddy:
@bodepd:  I didn't understand, why the conf  (cinder.conf) file changes go to upstream? These conf file changes are related to JioCloud specific, so no upstream needed here.

hkumarmk:
@swamireddy so basically all jiocloud specific configs/changes should be on puppet-rjil (unless it is not possible to be there in puppet-rjil at all)

swamireddy:
@hkumarmk: OK. Let me check if I can include it into puppet-rjil.

swamireddy:
please test this

swamireddy:
please test this

swamireddy:
please test this

swamireddy:
please test this

swamireddy:
please test this

swamireddy:
please test this

swamireddy:
please test this

swamireddy:
please test this

swamireddy:
Merge build failed with below error msg:
---
+ '[' '!' -f testjenkins-puppet-rjil-gate-1466_ssh_config ']'
+ ssh -F ssh_config -l jenkins gcp1_testjenkins-puppet-rjil-gate-1466 '~jenkins/tempest/run_tempest.sh -N -- --load-list ~jenkins/tempest_tests.txt'
ssh: Could not resolve hostname gcp1_testjenkins-puppet-rjil-gate-1466: Name or service not known
Build step 'Execute shell' marked build as failure
Finished: FAILURE
---
It says Could not resolve hostname gcp1_testjenkins-puppet-rjil-gate-1466: Name or service not known

is it known issue to be fixed??

swamireddy:
please test this

swamireddy:
please test this

swamireddy:
please test this

swamireddy:
please test this

swamireddy:
Thank you @sorenh. I have tested this functionality on the build Env and its working as expected.

PR #633, opened 2015-06-08
Avoid wan gossip by configuring consul with advertise_addr_wan localhost

hkumarmk:
We do not need wan nodes talk each other, but it seems wan gossip, causing the
nodes in different datacenters talk to each other, so setting advertise_addr_wan
to 127.0.0.1 in the environments which doesn't need connections between multiple
datacenters.


hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
This is not going to fix it as the nodes are joining to lan gossip and not to wan :( ,  so closing

PR #639, opened 2015-06-10
Added scripts to create admin_domain and cloud_admin

ajayaa:
This additionally adds the policy file for Keystone and modifies the policy file
to add the admin domain id to it.

Since Keystone v3 is not supported in icehouse version of puppet-keystone,
this script is created to create domain and admin for that domain.


manifests/keystone/cloud_admin.pp
```diff
@@ -0,0 +1,40 @@
+file { 'policy.json':
+  path    => '/etc/keystone/policy.json',
+  content => template('rjil/policy_json.erb'),
+  ensure  => 'file'
+} ->
```
hkumarmk: These file resources is not need to be in order, you can just remove '->' between file {}

manifests/keystone/cloud_admin.pp
```diff
@@ -0,0 +1,40 @@
...
+  content => template('rjil/cloud_admin_create.erb'),
+  ensure  => 'file',
+  mode    => 755
+} ->
+
+exec { 'cloud_admin_create':
+  path      => '/bin:/usr/bin:/usr/local/bin',
+  logoutput => true,
+  command   => 'echo "cloud_admin is not present";cloud_admin_create.sh',
+  unless   => 'cloud_admin_check.sh'
```
hkumarmk: you just need to add 
"require => [File['/usr/local/bin/cloud_admin_check.sh'],File['/usr/local/bin/cloud_admin_create.sh']]," 
so that this will execute after those file resources

manifests/keystone/cloud_admin.pp
```diff
@@ -0,0 +1,40 @@
...
+} ->
+
+exec { 'replace_domain_id':
+  path      => '/bin:/usr/bin:/usr/local/bin',
+  logoutput => true,
+  command   => 'replace_domain_id_in_policy_file.sh',
+  onlyif    => 'cloud_admin_check.sh'
```
hkumarmk: It seems this exec would always be running after cloud_admin_check.sh script land there, which is not puppet way of doing. It should only be running if the entry is not present in the file.

templates/cloud_admin_check.erb
```diff
@@ -0,0 +1,9 @@
+ADMIN_TOKEN=token123
```
hkumarmk: ADMIN_TOKEN should be a variable getting from the class, this should not be hardcoded.

templates/cloud_admin_check.erb
```diff
@@ -0,0 +1,9 @@
...
+AUTH_URL=https://identity.jiocloud.com:5000/v3
```
hkumarmk: same here, AUTH_URL should not be hardcoded

templates/cloud_admin_check.erb
```diff
@@ -0,0 +1,9 @@
...
+ADMIN_DOMAIN=$(curl -H "X-Auth-Token: $ADMIN_TOKEN" $AUTH_URL/domains?name=admin_domain | jq .domains[0].id | tr -d '"')
```
hkumarmk: what is this jq used for? is it available on the system?

templates/cloud_admin_create.erb
```diff
@@ -0,0 +1,56 @@
+ADMIN_TOKEN=token123
```
hkumarmk: same as above, admin_token, authurl should not be hardcoded

templates/cloud_admin_create.erb
```diff
@@ -0,0 +1,56 @@
...
+export ID_ADMIN_DOMAIN=$(\
+curl $AUTH_URL/domains \
+    -s \
+    -H "X-Auth-Token: $ADMIN_TOKEN" \
+    -H "Content-Type: application/json" \
+    -d '
+{
+    "domain": {
+    "enabled": true,
+    "name": "admin_domain"
```
hkumarmk: This should also be a variable which would be coming from the calling class, so it can be overridden

templates/cloud_admin_create.erb
```diff
@@ -0,0 +1,56 @@
...
+    -s \
+    -H "X-Auth-Token: $ADMIN_TOKEN" \
+    -H "Content-Type: application/json" \
+    -d "
+{
+    \"user\": {
+        \"description\": \"Cloud administrator\",
+        \"domain_id\": \"$ID_ADMIN_DOMAIN\",
+        \"enabled\": true,
+        \"name\": \"cloud_admin\",
```
hkumarmk: Should not be hardcoded

templates/cloud_admin_create.erb
```diff
@@ -0,0 +1,56 @@
...
+        \"password\": \"password\"
```
hkumarmk: must not be hardcoded

templates/policy_json.erb
```diff
@@ -0,0 +1,204 @@
+{
+    "admin_required": "role:admin",
+    "tenant_admin_required": "role:tenant_admin",
+    "cloud_admin": "rule:admin_required and domain_id:Default",
+    "service_role": "role:service",
+    "service_or_admin": "rule:admin_required or rule:service_role",
+    "owner" : "user_id:%(user_id)s or user_id:%(target.token.user_id)s",
+    "admin_or_owner": "(rule:admin_required and domain_id:%(target.token.user.domain.id)s) or rule:owner",
+    "tenant_admin_or_owner": "(rule:tenant_admin_required and domain_id:%(target.token.user.domain.id)s) or rule:owner or rule:cloud_admin",
+    "admin_or_cloud_admin": "rule:admin_required or rule:cloud_admin",
```
hkumarmk: isn't this rule same as admin_required?

templates/replace_domain_id_in_policy_file.erb
```diff
@@ -0,0 +1,8 @@
+ADMIN_TOKEN=token123
```
hkumarmk: same as above these things should not be hardcoded

templates/replace_domain_id_in_policy_file.erb
```diff
@@ -0,0 +1,8 @@
...
+AUTH_URL=http://identity.jiocloud.com:5000/v3
+
+ID_ADMIN_DOMAIN=$(curl -H "X-Auth-Token: $ADMIN_TOKEN" $AUTH_URL/domains?name=admin_domain | jq .domains[0].id | tr -d '"')
+echo "id_admin_domain is: $ID_ADMIN_DOMAIN"
+
+sed '/"cloud_admin": "rule:admin_required and domain_id:Default"/c\"cloud_admin": "rule:admin_required and domain_id:'"$ID_ADMIN_DOMAIN\"," /etc/keystone/policy.json > /tmp/policy.json
```
hkumarmk: better you make sure $ID_ADMIN_DOMAIN has values, otherwise it is easy to have blank value in policy.json in case of any glitch or anything.

templates/policy_json.erb
```diff
@@ -0,0 +1,204 @@
+{
```
hkumarmk: BTW this is a static file, so it doesnt need to be a template, this should move to files/ as keystone_policy.json or something like that (It should be preceded with keystone so it will not conflict with other policy.jsons. or it should be in a directory structure like files/keystone/policy.json

manifests/keystone/cloud_admin.pp
```diff
@@ -0,0 +1,40 @@
+file { 'policy.json':
+  path    => '/etc/keystone/policy.json',
+  content => template('rjil/policy_json.erb'),
```
hkumarmk: As the policy.json is a static file, it should be source => puppet:///modules/rjil/<pathto the file>

manifests/keystone/cloud_admin.pp
```diff
@@ -0,0 +1,40 @@
...
+  ensure  => 'file'
```
hkumarmk: ensure should be the first attribute (puppet-lint)

manifests/keystone/cloud_admin.pp
```diff
@@ -0,0 +1,40 @@
...
+} ->
+
+file { 'replace_domain_id_in_policy_file':
+  path    => '/usr/local/bin/replace_domain_id_in_policy_file.sh',
+  content => template('rjil/replace_domain_id_in_policy_file.erb'),
+  ensure  => file,
+  mode    => 755,
```
hkumarmk: mode should be 4 digit ('0755') and should be quoted (puppet-lint)

hkumarmk:
There is one basic mistake in this cloud_admin.pp file, this should be a class, so you should start with 

class rjil::keystone::cloud_admin ( 
   param1 => value,
   param2 => value2,
) {
  all code inside this block
}

hkumarmk:
Also you should include this class somewhere - It can be under rjil::keystone

hkumarmk:
Oh comments .... inline comments are always be helpful

manifests/keystone/cloud_admin.pp
ajayaa: But the operation is idempotent. So it should not be a problem.

templates/cloud_admin_check.erb
ajayaa: That needs to be installed wherever this code is running i.e. places where Keystone service is running. How do I install a package at on of the nodes?

templates/replace_domain_id_in_policy_file.erb
ajayaa: $ID_ADMIN_DOMAIN is being checked in the script cloud_admin_check.sh . This script runs only when that exits with 0 status.

ajayaa:
Agree that this things should be wrapped in a class some variables need to be made configurable.

templates/policy_json.erb
ajayaa: I don't understand what you are asking.

templates/replace_domain_id_in_policy_file.erb
hkumarmk: yeah, but what if the above curl call fail for some reason, which leaves that variable with no value which will be written in policy. so may be until next execution puppet execution, it may have blank value.

manifests/keystone/cloud_admin.pp
```diff
@@ -0,0 +1,52 @@
...
+    path    => '/etc/keystone/policy.json',
+    source => 'puppet:///modules/rjil/keystone_policy.json'
+  }
+
+  file { 'replace_domain_id_in_policy_file':
+    ensure  => 'file',
+    path    => '/usr/local/bin/replace_domain_id_in_policy_file.sh',
+    content => template('rjil/replace_domain_id_in_policy_file.erb'),
+    mode    => 755,
+  }
```
sorenh: This file contains some of the scariest credentials we have: The Keystone admin token. Mode 755 will make it readable by everyone.

sorenh: Besides, I recommend you make it '0755' (or whichever mode you decide on) like below. 755 decimal is 1363 in octal which is just bizarre.  I think Puppet does the right thing here, but it's a good thing to always ensure.

manifests/keystone/cloud_admin.pp
```diff
@@ -0,0 +1,52 @@
...
+    require   => [File['cloud_admin_check'], File['cloud_admin_create']]
+  } ->
+
+  exec { 'replace_domain_id':
+    path      => '/bin:/usr/bin:/usr/local/bin',
+    logoutput => true,
+    command   => 'replace_domain_id_in_policy_file.sh',
+    onlyif    => 'cloud_admin_check.sh',
+    require   => [File['cloud_admin_check'], File['policy.json'], File['replace_domain_id_in_policy_file']]
+  }
```
sorenh: So every puppet run will start by policy.json being put in place by the File resource at the top of this file, and then you'll replace some of the contents?

manifests/keystone/cloud_admin.pp
```diff
@@ -0,0 +1,52 @@
...
+  $admin_user_name     = 'cloud_admin',
+  $admin_user_password = 'password',
+  $admin_role          = 'admin',
+
+) {
+  file { 'policy.json':
+    ensure  => 'file',
+    path    => '/etc/keystone/policy.json',
+    source => 'puppet:///modules/rjil/keystone_policy.json'
+  }
```
sorenh: You should make the title of the resource bi the full path. That namespace is global. Another project wouldn't be able to create a file { 'policy.json': ... } resource.

manifests/keystone/cloud_admin.pp
sorenh: Oh, and you need to install the jq package. Just do it here in this file.

ajayaa: The middleware used to identify the admin token which is used to bootstrap Keystone should be disabled in Prod. That middleware is called admin_token_auth and it can be disabled through keystone-paste.ini. But afaik we don't have a way to manage paste file through puppet in Icehouse.

Have we disabled that middleware in our prod env?

manifests/keystone/cloud_admin.pp
ajayaa: done.

manifests/keystone/cloud_admin.pp
ajayaa: The scripts need to be run once to setup Keystone. But it won't be a problem if it is run multiple times also.
Do you know of a way to make sure it is run only once and that too successfully?

ajayaa:
please test this

ajayaa:
please test this

ajayaa:
please test this

ajayaa:
please test this

hkumarmk:
please test this

PR #640, opened 2015-06-10
Use old version of openstacklib to fix the failures

hkumarmk:
Changed .fixtures.yml to use jiocloud fork or openstacklib
    
    We were using openstacklib module from stackforge for travis, and openstacklib
    have changed the way they handle the openstack auth for openstack
    client based providers which affect us. Also we use jiocloud fork in Puppetfile.
    So changing it to jiocloud fork for travis also.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -73,7 +73,7 @@ fixtures:
       repo: https://github.com/stackforge/puppet-cinder
       ref: 4.0.0
     openstacklib:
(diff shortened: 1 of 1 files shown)


hkumarmk:
please test this

hkumarmk:
This is only change which affect travis test and it is passing. So merging this

mayankkapoor:
Support merging this to get builds to pass.

Issue #643, opened 2015-06-11
private_address and public_address is blank in compute nodes


hkumarmk:
This is happening because of we keep private and public_address to eth0/em1 etc, This will not work in compute node as compute node have the IPs transfered to vhost0.

As a temporary fix, there are custom facts to get IPs either vhost0 or ethx - this need to be used.

Because of this issue, cp30 in production is using a different IP address which it got for its first nic port.

PR #646, opened 2015-06-11
UC| Use interface and ip address based on the subnet

hkumarmk:
NOTE NOTE NOTE! THIS PATCH NEED CHANGES IN PRODUCTION HIERA (contrail::vrouter::vrouter_physical_interface to be added/changed)

Currently we assume that the ip address we need to configure in the cloud is
always setup on same interface on a specific set of servers/cloud provider,
which can be incorrect because of number of reasons
1. what if we had to use different nic port because of one bad port
2. compute node have the ip of physical interface will be moved to vhost0 in
which case you need special consideration - like custom facts

Rather the subnet is more consistent, this patch adds code to create custom
facts for ipaddress and interface based on subnet like interface_10_0_0_0_8 with
value eth0 in case eth0 have subnet configured 10.0.0.0/8.

This will allow us to use the fact ipaddress_{subnet} and interface_{subnet}
whereever required, as far as correct subnet provided, this will provide
consistant network config across the servers.

This should fix the issue #643.

### DIFF
--- a/hiera/data/cloud_provider/jio.yaml
+++ b/hiera/data/cloud_provider/jio.yaml
@@ -1,8 +1,3 @@
-public_address: "%{ipaddress_eth0}"
-public_interface: eth0
-private_address: "%{ipaddress_eth0}"
-private_interface: eth0
-
 rjil::system::proxies:
   "no":
     url: "127.0.0.1,169.254.169.254,localhost,consul,jiocloud.com"
--- a/lib/facter/jiocloud_network.rb
+++ b/lib/facter/jiocloud_network.rb
@@ -1,0 +1,32 @@
+##
+# Custom fact for ipaddress and interface based on network and netmask
+# This will help to configure ipaddress and interfaces for various services
+# without assuming IP address on specific interface.
+# If the machine have 2 nics with network 192.168.0.0/24 and 10.0.0.0/8, this
+# code will create facts like below with appropriate values.
+# ipaddress_192_168_0_0_24, interface_192_168_0_0_24,
+# ipaddress_10.0.0.0_24, interface_10.0.0.0_24
+##
+require 'ipaddr'
+Facter.value('interfaces').split(',').reject{ |r| r == 'lo' }.each do |iface|
+  ipaddress = Facter.value('ipaddress_' + iface)
+  if ipaddress
+    ##
+    # converting netmask to cidr, so that providing fact name would be easier
+    # - rather than typing ipaddress_192_168_0_0_255_255_255_0 just type
+    # 192_168_0_0_24 (<network>_<cidr>)
+    ##
+    cidr = IPAddr.new(Facter.value('netmask_' + iface)).to_i.to_s(2).count("1").to_s
+    network = Facter.value('network_' + iface).gsub('.', '_')
+    Facter.add('ipaddress_' + network + '_' + cidr) do
+      setcode do
+        ipaddress
+      end
+    end
+    Facter.add('interface_' + network + '_' + cidr) do
+      setcode do
+        iface
+      end
+    end
+  end
+end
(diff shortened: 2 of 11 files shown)


lib/facter/jiocloud_network.rb
```diff
@@ -0,0 +1,32 @@
...
+Facter.value('interfaces').split(',').reject{ |r| r == 'lo' }.each do |iface|
+  ipaddress = Facter.value('ipaddress_' + iface)
+  if ipaddress
+    ##
+    # converting netmask to cidr, so that providing fact name would be easier
+    # - rather than typing ipaddress_192_168_0_0_255_255_255_0 just type
+    # 192_168_0_0_24 (<network>_<cidr>)
+    ##
+    cidr = IPAddr.new(Facter.value('netmask_' + iface)).to_i.to_s(2).count("1").to_s
+    network = Facter.value('network_' + iface).gsub('.', '_')
```
sorenh: Is this gsub necessary? Can we not keep the dots?

sorenh: I just think address_10.0.0.0_16 is much easier on the eyes.

hkumarmk: but puppet doesn't seem to be happy to have dot in variables

hkumarmk:
please test this

lib/facter/jiocloud_network.rb
sorenh: Ok. :cry:

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
@bodepd @sorenh could you please review it. I have triggered a test build, if all goes well, I would like to get it merged - so I dont have to create another set of facts for undercloud prestaging work.

anshprat:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hiera/data/cloud_provider/jio.yaml
```diff
@@ -1,8 +1,3 @@
-public_address: "%{ipaddress_eth0}"
```
bodepd: I'm a little concerned about now support the old interface based model b/c it adds in the constraint that networks have to have exactly a 10.0.0.0/16. So if anyone tries to deploy against a different subnet, things will fail in a non-obvious way. Is this being accounted for?

bodepd: after reading through the issues and commit about this, I can see why it's required, but it's only really required for physical and not virtual interfaces.

bodepd:
If I merge this, how are you going to coordinate to make sure the corresponding changes are made in prod?

bodepd:
I've reviewed this patch, verified that it works as expected. I have two concerns that I outlined above:

1. it places constraints on virtual environments that don't need to exist (about hard-coded network CIDRs,)
2. It also a little concerned what happens if the fact does not return anything (in the case where that network doesn't match).

It tried setting it to something invalid, and saw that a catalog still got compiled. In my case of ocdb, it just happened to fail during run-time evaluation, but only b/c of code from the db:

````
Jun 29 04:55:20 ocdb1-testjenkins-puppet-rjil-gate-1617 puppet-user[27450]: Compiled catalog for ocdb1-testjenkins-puppet-rjil-gate-1617.node.consul in environment production in 22.42 seconds
Jun 29 04:55:22 ocdb1-testjenkins-puppet-rjil-gate-1617 puppet-user[27450]: Parameter name failed on Mysql_user[monitor@]: Invalid database user monitor@ at /etc/puppet/modules.overrides/rjil/manifests/db.pp:115
````

If we failed much more explicitly, it would protect against the case where a CIDR is set incorrectly, and also against prod getting hosed and all ips, interfaces getting unset.

hkumarmk:
please test this

hkumarmk:
@bodepd I made the change you suggested, please take a look at this.

hkumarmk:
@bodepd Please take a look at this. This is required for UC

hkumarmk:
please test overcloud

hkumarmk:
please test overcloud

hkumarmk:
please test overcloud

PR #648, opened 2015-06-12
Adding role specific override per environments

hkumarmk:
Currently if we wanted to override things per role in specific environment, it
is not possible (say i need to override a setting on ct role in production). So
adding a role specific override per environments.

### DIFF
--- a/hiera/hiera.yaml
+++ b/hiera/hiera.yaml
@@ -7,6 +7,7 @@
 :hierarchy:
   - "clientcert/%{::clientcert}"
   - "secrets/%{env}"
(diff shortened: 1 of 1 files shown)


bodepd:
I'm a little worried about this feature as a way to make things more different between envs than we want them to be (which can often be a cause of failures).

do you mind saying the specific case?

hkumarmk:
one case may be say in production, we have internet is on gcp1 where we can run ntp instead of bootstrap. 

In fact any role based overrides in specific environment should need this.

hkumarmk:
In fact this patch was created because of a reason where we had to create a user only on one role.

PR #651, opened 2015-06-15
WIP | Undercloud controller work

hkumarmk:


hkumarmk:
please test this

manifests/openstack_objects.pp
```diff
@@ -10,6 +10,7 @@
   $override_ips      = false,
   $users             = {},
   $tenants           = undef,
+  $lb_available      = true,
```
bodepd: please submit this as a separate patch that we can test individually.

bodepd:
I took a quick look at the code, and I have the following comments.

* Since it's such a large patch, I'd like to do the following:
  * split up all parts that might effect the existing builds so that we can validate and merge those individually
  * submit the uc specific parts as a single patch that doesn't touch any existing code so that we can just run a test and merge it without risk to the existing overcloud stuff

Also, can we set up gate tests for undercloud? It would be great to do that as a part of this process so that we can continuously test that code, otherwise, it will just stop working.

hkumarmk:
Sure, thats what I was also thinking - I wanted to make it work first and then will split in different PRs.

hkumarmk:
please test me

hkumarmk:
please test me

hkumarmk:
please test me

hkumarmk:
please test this

hkumarmk:
please test me

hkumarmk:
There are problems with environments - sometimes, nova boot causing errors, sometimes quota issues :(

hkumarmk:
please test me

hkumarmk:
please test me

hkumarmk:
please test me

hkumarmk:
please test me

hkumarmk:
please test me

templates/tests/neutron.sh.erb
```diff
@@ -1,21 +1,25 @@
...
-  eval $cleanup_command
   echo "CRITICAL: $@"
   exit 2
 }
 
 if [ -f /root/openrc ]; then
   source /root/openrc
-  netname=`hostname`
   neutron net-list -D || fail 'neutron net-list failed'
+  <% if @test_netcreate -%>
```
bodepd: what's the point of even having these tests here if they are not going to create test networks?

templates/tests/neutron.sh.erb
```diff
@@ -1,21 +1,25 @@
...
+  netname=`hostname`
+  for net in `neutron net-list -D | grep $netname | awk '/[a-z][a-z]*[0-9][0-9]*/ {print $2}'`; do
+    echo "Found unexpected network leftover from previous test"
+    neutron net-delete $net || fail 'neutron net-delete failed'
```
bodepd: won't this fail if the ports also exist? Can you delete networks that have attached ports?

manifests/test/neutron.pp
```diff
@@ -2,8 +2,20 @@
 # Class: rjil::test::neutron
 #   Adding tests for neutron services
 #
+# == Parameters
```
bodepd: this can be reasonably split out into its own patch.

manifests/neutron/network.pp
```diff
@@ -0,0 +1,48 @@
+##
+# Create network and subnet
+##
+
+define rjil::neutron::network (
```
bodepd: is this just used to the undercloud controller?

build_scripts/deploy.sh
```diff
@@ -8,6 +8,8 @@ then
     consul_discovery_token=$(curl http://consuldiscovery.linux2go.dk/new)
 fi
 
+export $consul_discovery_token
```
bodepd: why does this have to be exported?

bodepd:
I really just see two patches here that need to be spit out.

1. patch to test the no mk_resource_methods branch of neutron (I remember that this failed last time that I test it)
2. neutron test scripts

templates/tests/neutron.sh.erb
hkumarmk: Even though it cannot create/delete test networks, it can list net-list, so it make sense to be there.

templates/tests/neutron.sh.erb
hkumarmk: You cannot delete any networks which have attached ports, but this is just deleting only the test networks which dont have any ports attached. BTW, since it only happen to be if test_netcreate set, it is not going to be there in uc.

manifests/test/neutron.pp
hkumarmk: I can split this

manifests/neutron/network.pp
hkumarmk: I split this part to common code, so it can be used anywhere else, currently it is used only for undercloud controller.

build_scripts/deploy.sh
hkumarmk: This is because overcast is handling it within it, in separate shell, which would not be available without exporting to environment.

PR #655, opened 2015-06-15
vrouter restart

anshprat:


sorenh:
Even if the lock_file exists, you should still perform the check. If the check succeeds, delete the lock file and return success. If it still fails, just exit 1.

sorenh:
Maybe this can be done even more easily: How about this?

Perform the check
If it succeeds, delete any lockfile, exit 0
If it fails, check for lock file.
If lock file exists, exit 2.
If lock file does not exist, create it, issue the restart, exit 1.

Done. The first time it sees the stuck agent, it restarts and sets the service to WARNING. The next time it runs (around 10 seconds later) it ends up in CRITICAL if it still doesn't work and OK if it recovered.

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

anshprat:
please test this

bodepd:
please test this

bodepd:
I've reviewed this patch, it does exactly what it is supposed to, it works, doesn't have much risk. Integrated correctly with consul.

From an implementation perspective it's ready to be merged. I am hesitant to merge it b/c of the "should we merge it question" (ie: is this the approach that we want at the minute, etc.)

@mayankkapoor @hkumarmk  @sorenh can one of you guys decide and merge it?

PR #659, opened 2015-06-16
ceph logrotate

anshprat:


anshprat:
please test this

bodepd:
more context in the commit message would be great. Especially about that post script and what it does.

Also, this build looked ok when I logged in, it was just missing a node

bodepd:
please test this

bodepd:
I have a passing build on 1514 (49.40.64.7) , all in all, looks good, I would just like to see it in action (ie: wait 24 hours to see rotation work), also, I need more context on what that postscript does, this info should be in the commit message.

anshprat:
@bodepd - I just copied the postrotate from existing packaged / upstream logrotate, hence the lack of comment in commits ( I did add a note about it being from upstream in the post-rotate file). But basically the post rotate file just does a reload. The reason for a separate post-rotate file here is that depending upon the ceph conf, a generic service/system ceph reload could be a no-op. So the post-rotate script tried to do a reload in a generic way (service/invoke-rc.d) and then using some ceph specific lock files + initctl. This could possible cause double reload (as mentioned in the reload script) but is basically to handle a possible no-op in the first reload.   
To reiterate - its a copy from upstream and since it was already working in prod, I did not make any change to post-rotate script (for example in the second reload we could possible check the age of ceph before trying reload). The only change in this logrotate PR is increase the log retention to 60 from 7 (+ puppetization).

bodepd:
please provide as much context as you can in the commit message or as an inline comment. No problem with the code, I just want to see if after 24 hours.

bodepd:
I left a build up for a day and observed that it did create:

is this the format you expected instead of .N? If so, it's a little inconsistent with the way it was done on other nodes:

````
ceph-mon.stmonleader1-testjenkins-puppet-rjil-gate-1514.log-20150618
````

I also noticed two other things:

1. radowgw was not rotated (it does not end in .log), was this intentional?
2. maybe-upgrade is completely stuck on the radosgw check (it's been stuck for over a day). Soren was doing some testing, so it might be related to that.

I will rerun a build to see if the stuck maybe-upgrade script happens again on stmons

bodepd:
please test this

anshprat:
@bodepd 

1) about the date format - yes, its expected. And I expect the same for all logs. This is one of the discrepancies I noticed yesterday.

The default jiocloud/logrotate.pp has  
`  $dateext       = true,`
So, I expect all the logs to have the datestamp, but for some reasons its not happening for all nodes. 

2) radosgw not being rotated - no, its not intentional. Looks like its not at all managed presently. Will add it.

anshprat:
@bodepd https://github.com/JioCloud/puppet-ceph/pull/17 will fix the radosgw log.

anshprat:
please test this

anshprat:
please test this

PR #662, opened 2015-06-17
Sudo with different capabilities

hkumarmk:
This patch enable to create sudo users with different capabilities, It
    support adding hiera data with different set of users with different commands
    allowed.
    
    e.g below hiera data will create two sets of users, - admin1 and admin2 with all
    access and user1 and user2 with access to commands cat and less.
    ```
    rjil::system::accounts::sudo_users:
      admin:
        users: [admin1,admin2]
      readonly:
        users: [user1,user2]
        commands_allowed: ['/bin/cat','/usr/bin/less']
    ```
    
    Appends username to command alias to prevent errors when there are multiple
    non-admin users.

### DIFF
--- a/hiera/data/env/at.yaml
+++ b/hiera/data/env/at.yaml
@@ -8,7 +8,9 @@ rjil::ceph::osd::osd_journal_size: 2
 nova::compute::libvirt::libvirt_virt_type: qemu
 
 rjil::system::accounts::active_users: [soren,bodepd,hkumar,jenkins,consul,pandeyop,jaspreet,vivek,ahmad,vaidy,himanshu,rohit,sanjayu,alokjani,amar,ynshenoy,abhishekl,mayankkapoor,soumit,anshup,varunarya,prashant,punituee,saju,ajayaa,swami,karansa,abhidixit,shashankchakelam,vpramo,atmesh.mishra,prashant.chawla]
-rjil::system::accounts::sudo_users: [soren,bodepd,hkumar,jenkins,consul,pandeyop,jaspreet,vivek,ahmad,vaidy,himanshu,rohit,sanjayu,alokjani,amar,ynshenoy,abhishekl,mayankkapoor,soumit,anshup,varunarya,prashant,punituee,saju,ajayaa,swami,karansa,abhidixit,shashankchakelam,vpramo,atmesh.mishra,prashant.chawla]
+rjil::system::accounts::sudo_users:
+  admin:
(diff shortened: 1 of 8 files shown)


hkumarmk:
please test this

hkumarmk:
This is working, I tested with two users - hkumar and sanjayu on bootstrap1 of the build 1524.

The build failure is because of ntpdate failure - I see this behavior on multiple builds.

@sorenh @bodepd please take a look at it, and if good, I wanted to get this merged asap as this is required to reach production before we can provide read access to engineering people.

hiera/data/env/at.yaml
```diff
@@ -8,7 +8,12 @@ rjil::ceph::osd::osd_journal_size: 2
 nova::compute::libvirt::libvirt_virt_type: qemu
 
 rjil::system::accounts::active_users: [soren,bodepd,hkumar,jenkins,consul,pandeyop,jaspreet,vivek,ahmad,vaidy,himanshu,rohit,sanjayu,alokjani,amar,ynshenoy,abhishekl,mayankkapoor,soumit,anshup,varunarya,prashant,punituee,saju,ajayaa,swami,karansa,abhidixit,shashankchakelam,vpramo,atmesh.mishra,prashant.chawla]
-rjil::system::accounts::sudo_users: [soren,bodepd,hkumar,jenkins,consul,pandeyop,jaspreet,vivek,ahmad,vaidy,himanshu,rohit,sanjayu,alokjani,amar,ynshenoy,abhishekl,mayankkapoor,soumit,anshup,varunarya,prashant,punituee,saju,ajayaa,swami,karansa,abhidixit,shashankchakelam,vpramo,atmesh.mishra,prashant.chawla]
+rjil::system::accounts::sudo_users:
+  admin:
+    users: [soren,bodepd,jenkins,consul,pandeyop,jaspreet,vivek,ahmad,vaidy,himanshu,rohit,sanjayu,alokjani,amar,ynshenoy,abhishekl,mayankkapoor,soumit,anshup,varunarya,prashant,punituee,saju,ajayaa,swami,karansa,abhidixit,shashankchakelam,vpramo,atmesh.mishra,prashant.chawla]
+  readonly:
+    users: [hkumar]
```
sorenh: Is this intentional? You only want readonly rights in AT?

hkumarmk: Oh I just made it to test it, I will just revert that change :)

hiera/data/env/at.yaml
```diff
@@ -8,7 +8,12 @@ rjil::ceph::osd::osd_journal_size: 2
...
+    commands_allowed: [/bin/cat,/usr/bin/less]
```
sorenh: This is a security problem. less lets you execute arbitrary commands (with the '!' command) as well as switch to edit mode (try pressing 'v').

hkumarmk:
@sorenh Done, reverted the change to the user hkumar

hiera/data/env/at.yaml
hkumarmk: oh yeah :) I will just make it cat they can pipe to any pager they want

hkumarmk: oh cat also may be a bit problem as that can be used to write the files

hkumarmk: may be view?

PR #664, opened 2015-06-17
rjil::test::check changes

hkumarmk:
This patch:
* Added a parameter to rjil::test::check to use it for both service and
 validation checks
* Added a udp port check script template using netcat - nagios check_udp doesnt
 work as expected

This would be using in undercloud controller, just pushing it as separate pr, if we wanted to test it separately.

### DIFF
--- a/manifests/test/check.pp
+++ b/manifests/test/check.pp
@@ -4,15 +4,24 @@
 #
 
 define rjil::test::check(
-  $port    = 0,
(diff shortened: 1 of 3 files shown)


bodepd:
@hkumarmk I really want to start seeing unit tests on changes like this.

hkumarmk:
@bodepd done, added spec tests

bodepd:
please test this

Issue #663, opened 2015-06-17
some builds are failing because of ntpdate doesnt work on bootstrap1


hkumarmk:
It seems ntpdate to pool.ntp.org is not working sometimes which is causing the build failures. 

I see a solution to run ntpdate on all but bootstrap1 which is the local ntp server for the build or may be we could sync the time with any other internal node (may be httpproxy node?).

bodepd:
what changed that is causing this, what was the symptom? I haven't been seeing this.

hkumarmk:
I have seen at least 3 times from yesterday and today in different builds - 2 in undercloud builds and one on gates - the symptom is ntpdate (to pool.ntp.org) just dont work on bootstrap and thus ntp.service.consul never get registered in consul which is causing build failures.

anshprat:
resolved via PR, closing.

PR #665, opened 2015-06-17
Disable execution of ntpdate on ntp server

hkumarmk:
Sometimes ntpdate to pool.ntp.org is keep on failing, which is causing the
entire build failing, This patch is to disable ntpdate on local ntp server which
should not cause any issues as other nodes are syncing the time from this
server and until ntp server get the time from remote clock, it will use local
clock, so still all nodes in the build will have clock synced.

This solve issue #663.

### DIFF
--- a/hiera/data/role/bootstrap.yaml
+++ b/hiera/data/role/bootstrap.yaml
@@ -1,6 +1,7 @@
 rjil::jiocloud::consul_role: bootstrapserver
 # etcd serves as the ntp server
 rjil::system::ntp::server: true
+rjil::system::ntp::run_ntpdate: false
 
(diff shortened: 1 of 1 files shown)


hkumarmk:
please test this

bodepd:
please test this

bodepd:
I ran a build with this and ntpdate never got called b/c it doens't get called if ntp is already running. Are you expecting it to get called?

hkumarmk:
@bodepd yeah, but it getting called before ntp server setup on bootstrap, and if ntpdate (to external server) failed, the entire build will be failing as ntp.service.consul will not be created without successful ntpdate.

PR #670, opened 2015-06-22
Added hkumar to sudo  admin list

hkumarmk:
This was missed before, so adding myself in sudo admin list.


hkumarmk:
merging it since it will not affect anything else.

PR #673, opened 2015-06-22
Reduced footprint for the builds

hkumarmk:
We have the flavors with good amount of resources, so we do not need to use
large flavors for our builds, so changing as below
medium to small
large to mmedium
storage to medium

### DIFF
--- a/environment/jio.map.yaml
+++ b/environment/jio.map.yaml
@@ -2,9 +2,9 @@ image:
   trusty: 5d246189-a666-470c-8cee-36ee489cbd9e
 flavor:
   small: da9ba7b5-be67-4a62-bb35-a362e05ba2f2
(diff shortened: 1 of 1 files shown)


hkumarmk:
please test this

hkumarmk:
please test this

sorenh:
On HP Cloud, the largest instance type we used was 102. I believe they have 4 GB of RAM which is pretty close to our "small". I think using "small" for everything might work.

hkumarmk:
oh okay, lemme try with small then (better to start with smallest flavor :) )

hkumarmk:
please test this

sorenh:
YESYYESYEYSYES!!!!111!!!!!!!eleven!

PR #674, opened 2015-06-23
Update tempest configuration file for gate environments

pandeyop:
There are changes to tempest configuration file so as to run all tempest tests on the gate environments. One of the restrictions is to boot only from volume in addition to other configurations.


pandeyop:
please test this

PR #676, opened 2015-06-23
DNS update for getting public IP for api.jiocloudservices.com

vpramo:
Override for dnsmasq in compute nodes to ensure dnsmasq does not server request from /etc/hosts file in compute nodes


vpramo:
please test this

hiera/data/role/cp.yaml
```diff
@@ -0,0 +1,2 @@
+#Adding dnsmasq override for GCP node
```
hkumarmk: not gcp node, cp node

PR #679, opened 2015-06-23
DNS update for getting public IP for api.jiocloudservices.com

vpramo:
Override for dnsmasq in compute nodes to ensure dnsmasq does not server request from /etc/hosts file in compute nodes


sorenh:
I'm not sure about this.. If we don't want dnsmasq to use /etc/hosts, why are we putting these entries in there to begin with?

sorenh:
Sorry, that of course makes sense.

However, I'm fairly sure it won't fix the problem. To the best of my knowledge, Contrail does a getaddressinfo(), so it'll just as happily use a value from /etc/hosts as any other application would.

vpramo:
please test this

vpramo:
@sorenh Ok, i discussed this with Harish, let me check this out. If this will not work, then maybe adding a nameserver to 8.8.8.8 maybe the only solution

hkumarmk:
@sorenh I thought vrouter make  a dns query in datacenter infrastructure from compute node :( if it does getaddressinfo(), that would be a problem. In that case we would need some different things.

Anyway even if it fix the current problem, there are other major issue (of internal dns names resolved by vms) which would need a different fix to solve.

sorenh:
I have an idea.

Create /etc/contrail-resolv.conf that simply reads "nameserver <ip of desired nameserver>"
Create /etc/contrail-hosts that just has localhost entries.

Now change contrail-vrouter-agent's upstart job to do:
unshare -m /bin/sh <<EOF
mount -o bind /etc/contrail-resolv.conf /etc/resolv.conf
mount -o bind /etc/contrail-hosts /etc/hosts

# ...and then add all the stuff from the existing script in contrail-vrouter-agent's upstart job
EOF

This creates a new mount namespace just for contrail-vrouter-agent. Then we can mount replacement resolv.conf and hosts over it without affecting anything else on the system.

sorenh:
@vpramo Can you take my suggestion and implement it and see if it works as expected?

vpramo:
@sorenh  Sure i am trying it out on the build, will update you today

vpramo:
@sorenh  I tried it out, i made the edits and restarted the agent service , the service is no longer visible  in ps and after which locally i cannot ssh to my vm using the link local address.

sorenh:
Because the original script from the upstart job now is embedded in a heredoc, we need to escape the $'s.

sorenh:
So the script becomes:

    script
      unshare -m /bin/sh <<EOF
      mount -o bind /etc/contrail-resolv.conf /etc/resolv.conf
      mount -o bind /etc/contrail-hosts /etc/hosts
      COMMAND="/usr/bin/contrail-vrouter-agent"
      CONF="/etc/contrail/contrail-vrouter-agent.conf"
      USER="contrail"
      OPTS="--config_file \${CONF}"

      # Allow override of command/conf and opts by /etc/default/daemon-name
      if [ -f /etc/default/\$UPSTART_JOB ]; then
        . /etc/default/\$UPSTART_JOB
      fi

      if ! [ -r "\$CONF" ] ; then
        echo "Could not read \${CONF}: exiting"
        exit 0
      fi

      exec start-stop-daemon --start \
            --pidfile /var/run/\${UPSTART_JOB}.pid \
            --exec \$COMMAND -- \$OPTS
      EOF
    end script

vpramo:
please test this

vpramo:
@sorenh @hkumarmk @bodepd  I have finally incorporated the init script override in puppet-rjil. Please review the same. The init script is generally packaged as part of the vrouter package, hence i am not sure if replacing the file from puppet-rjil is a permanent solution

manifests/contrail/agentrewrite.pp
```diff
@@ -0,0 +1,42 @@
+# Class to rewrite the contrail-vrouter-agent.conf file
```
hkumarmk: Why do we need this file? This could be added in vrouter.pp itself.

manifests/contrail/vrouter.pp
```diff
@@ -48,6 +48,13 @@
     logdir       => '/var/log/contrail',
     copytruncate => true,
   }
+
+  # Class to overwrite the /etc/init/contrail-vrouter-agent.conf
+  # so that the agent can have a private namespace mount point of
+  # /etc/hosts and /etc/resolv.conf. This fix is done so that VM's 
+  # cannot resolve internal IP's of the cloud
+
+  include rjil::contrail::agentrewrite
```
hkumarmk: Better you added the code in this class to rjil::contrail::vrouter

hkumarmk:
I think in this case, we dont need bunch of code which does changing dnsmasq to not to read /etc/hosts etc.

Here is what I think we can do:

Create /etc/contrail-resolv.conf - make it as a template and get the nameserver[s] from the param - this param should be the upstream caching dns server for VMs. So contrail-vrtouer agent will not have any idea about any internal dns names (including *.consul) and thus the VMs running on compute node.

files/contrail-resolv.conf
```diff
@@ -0,0 +1,5 @@
+# HEADER: This file was autogenerated by puppet.
+# HEADER: While it can still be managed manually, it
+# HEADER: is definitely not recommended.
+nameserver 127.0.0.1
```
hkumarmk: This should be a template and should get name server[s] ip from the class param.

files/contrail-vrouter-agent.conf
```diff
@@ -0,0 +1,33 @@
+description "Contrail Vrouter Agent"
```
hkumarmk: As you said, it should be coming from the package, @sorenh ?

hiera/data/role/cp.yaml
```diff
@@ -0,0 +1,2 @@
+#Adding dnsmasq override for CP node
+rjil::jiocloud::dnsrewrite::content: 'no-hosts'
```
hkumarmk: This is not required, if we configure the upstream(only for vms) caching dns server in /etc/resolv.conf

manifests/contrail/agentrewrite.pp
```diff
@@ -0,0 +1,42 @@
...
+# to create a private namespace for mounting /etc/contrail-hosts
+# and /etc/contrail-resolv.conf on  to /etc/hosts and /etc/resolv.conf
+class rjil::contrail::agentrewrite{
+
+  include contrail::vrouter
+
+
+  file {'/etc/contrail-hosts':
```
hkumarmk: Not required if we NOT using dnsmasq (just use upstream caching name server)

files/contrail-vrouter-agent.conf
```diff
@@ -0,0 +1,33 @@
...
+start on (local-filesystems and net-device-up IFACE!=lo)
+stop on runlevel [!2345]
+
+chdir /var/run
+respawn
+
+script
+  unshare -m /bin/sh <<EOF
+  mount -o bind /etc/contrail-resolv.conf /etc/resolv.conf
+  mount -o bind /etc/contrail-hosts /etc/hosts
```
hkumarmk: Not required, just use upstream caching name server IP in above contrail-resolv.conf

manifests/contrail/vrouter.pp
```diff
@@ -48,6 +48,13 @@
     logdir       => '/var/log/contrail',
     copytruncate => true,
```
hkumarmk: Add nameservers param (may be taking an array), which should be used in contrail-resolv.conf.erb template - keep localhost as default,

manifests/jiocloud.pp
```diff
@@ -81,4 +82,10 @@
     setting => 'manifestdir',
   }
 
+
+  # Adding override configuration for dnsmasq on all compute nodes to
+  # ensure dnsmasq does not pick Cnames from /etc/hosts. This will enable
+  # any prod VM's to resolve api.jiocloudservices.com to public IP
+  include rjil::jiocloud::dnsrewrite
```
hkumarmk: Not required

manifests/jiocloud/dnsrewrite.pp
```diff
@@ -0,0 +1,10 @@
+class rjil::jiocloud::dnsrewrite($content='#no-hosts'){
```
hkumarmk: This is not required

hkumarmk:
BTW,  I confirmed that vrouter is not calling any internal names like *.consul or api.jiocloud* so disabling access to internal dns (including consul dns) should not affect vrouter functionality.

@sorenh This solution is great!! It really solve all those security problems because of internal dns being accessed by vms :+1:

sorenh:
@hkumarmk Oh, right, I hadn't considered that vrouter-agent might have legitimate uses for consul based names. Lucky that it doesnt!

files/contrail-vrouter-agent.conf
sorenh: What's not required? The /etc/hosts override?

hkumarmk: oh yeah sorry that is required (that time I just thought that it was only dns thing :) )

vpramo:
@sorenh  Yes its not required, as we will ensure that /etc/contrail-resolv.conf gets pointed to GCP node in production and keep it open in AT

manifests/contrail/agentrewrite.pp
hkumarmk: sorry it is required as vrouter also may resolve the names in /etc/hosts also for vms

hkumarmk:
@sorenh yeah we are resolving it in puppet and writing IPs in vrouter.conf

files/contrail-vrouter-agent.conf
sorenh: Yeah, we either need to override /etc/hosts to not have special entries or override nsswitch.conf to skip the /etc/hosts lookup altogether. /etc/hosts seems most straightforward.

hkumarmk: Oh.. no, we just tested it - vrouttr is not exposing the IPs in /etc/hosts to vms :) - We put say an ip in /etc/hosts for google.com and tried to get it resolved from the vms, and it was not.

So this is NOT needed to have /etc/hosts to be overridden

vpramo:
@sorenh  Can we package the modified init script as part of the vrouter package. It will be better that way right or shall we keep it as an override functionailty

vpramo:
Please test this

vpramo:
@sorenh and @hkumarmk Please review , the changes.

vpramo:
please test this

sorenh:
I considered whether this should go into the contrail packages, but since it'll be referring to files that aren't in the package (namely the override files), that seems inconsistent.

I think we should file a bug upstream about this, though. I can't imagine we're the only ones who need to sanitize DNS access from VM's.

hiera/data/env/at.yaml
```diff
@@ -37,3 +37,5 @@ rjil::neutron::contrail::fip_pools:
     tenants: ['tempest','mgmt_tenant','internal_tenant']
 
 contrail::vrouter::vgw_subnets: ['100.1.0.0/24']
+
+rjil::contrail::vrouter::dns_nameserver_list: ['127.0.0.1','127.0.1.1']
```
sorenh: There's no reason to list localhost twice. It will only double the time to timeout if it's down.

sorenh:
Fix the double listing of localhost in resolv.conf, and it's good to merge.

hiera/data/env/at.yaml
vpramo: @sorenh this is for testing environment, for production implementation i guess you or Harish have to add the variable to production.yaml with the IP address of the GCP node. 

This i think will affect only the builds i guess. I will correct it and resubmit it tomorrow if required.

hkumarmk: In fact this hiera is not required at all for testing environment, as this is default. We will add this for production or whereever required.

vpramo: @hkumarmk So can this be merged and may i know when this will go into production, i can then tell the object storage team to test it out.

hiera/data/role/gcp.yaml
```diff
@@ -6,3 +6,4 @@ contrail::vrouter::vgw_enabled: true
 contrail::vrouter::dest_net: 0.0.0.0/0
 contrail::vrouter::vrf: 'default-domain:services:public:public'
 rjil::test::contrail_vrouter::vgw_enabled: "%{hiera('contrail::vrouter::vgw_enabled')}"
+
```
hkumarmk: This is not necessary. Please remove this change

hiera/data/env/at.yaml
```diff
@@ -37,3 +37,4 @@ rjil::neutron::contrail::fip_pools:
     tenants: ['tempest','mgmt_tenant','internal_tenant']
 
 contrail::vrouter::vgw_subnets: ['100.1.0.0/24']
+
```
hkumarmk: This is not necessary, please remove this newline

manifests/jiocloud.pp
```diff
@@ -13,6 +13,7 @@
 
   include rjil::system::apt
 
+
```
hkumarmk: These new lines are not necessary.

hkumarmk:
@vpramo I commented to remove some newlines which was the only change on those files, please remove them.

Also please cleanup the history - it is not required 11 commits for this small change.

vpramo:
please test this

vpramo:
@sorenh and @hkumarmk  Please review and merge. Also please update the heira variable in production.yaml

vpramo:
please test this

hkumarmk:
@vpramo merging this, it would be great if you could write spec tests also in future.

manifests/contrail/vrouter.pp
```diff
@@ -48,6 +49,42 @@
...
   }
+
+  # overwrite the /etc/init/contrail-vrouter-agent.conf
+  # so that the agent can have a private namespace mount point of
+  # /etc/hosts and /etc/resolv.conf. This fix is done so that VM's 
+  # cannot resolve internal IP's of the cloud
+  include contrail::vrouter
```
hkumarmk: hey why you need to import this? This is already being called from the same class. This is not required.

hkumarmk:
@vpramo please fix the above change too, so your patch can be merged. sorry I could not catch this before.

vpramo:
please test this

vpramo:
@sorenh and @hkumarmk  Done with the changes , please review and merge

PR #681, opened 2015-06-23
Added an user in mysql with readonly access

amar266:
Added an user in mysql with readonly access to all databases


bodepd:
More context as to the reason this user is being created would be appreciated.

amar266:
I got a request that the engineering team needs to view the databases for debugging, for that we need to create an user  with readonly access to databases.

bodepd:
@mayankkapoor I'm a little worried about this patch b/c it requires a password with access to dbs to be floating around. Is there a way that we do this that is more secure? Like could we create a user, with auth_keys, and configure it to use this user, so that distributing a password doesn't have to be a part of this process?

@sorenh @hkumarmk thoughts?

hkumarmk:
it is doable by creating .my.cnf file with appropriate details under user home directory.

amar266:
please test this

amar266:
@hkumarmk @bodepd @sorenh  if i create  .my.cnf with username and password to every users home directory , then i will not have to distribute username and password to everybody.If we agree on this , i will make the code changes.

hkumarmk:
:+1:

sorenh:
What about a special "dbreadonly" user who has this my.cnf. Then we can add something like this to sudoers:
%engineering ALL = (dbreadonly) /usr/bin/mysql

That would let everyone in the engineering group do "sudo -u dbreadonly mysql". That seems easier than having to put a file in everyone's home dir. They can still get to the credentials (the mysql client lets you read files from the local disk as well as execute arbitrary commands), but it seems like a step in the right direction.

hkumarmk:
@sorenh yeah, do you see any other benefit on creating a special user?

Because implementation perspective, creating a file would not be that difficuilt, as we could conditionally create a file{} resource in manifests/system/accounts/localuser.pp?

but creating sudo with different "as user (dbreadonly vs root)" for same set of people or creating groups and adding them in the groups would also need changes in the code. (btw, It may be easier to do group management for future)

sorenh:
Hm. Ideally, we don't want to expose credentials and this seems like a step in the right direction, but given the capabilities of the mysql client that I outlined in my previous comments, it doesn't really matter. The only other reason I can think of is that it's fewer resources for Puppet to manage, but it's a small number, so again: It doesn't really matter.

Do it whichever way you prefer.

PR #682, opened 2015-06-23
UC|Create common code for network interface configuration

hkumarmk:
Network interface configuration code need to be reused. I checked if we could
use the module puppet-network, but it dont have the configuration which we need.
Another reason is that we would need to refresh the network before writing the
configuration, which is not handled in the existing code in the module.

Added spec tests also

### DIFF
--- a/files/etc_network_interfaces
+++ b/files/etc_network_interfaces
@@ -1,0 +1,11 @@
+##
+# THIS FILE IS MANAGED BY PUPPET
+##
+# This file describes the network interfaces available on your system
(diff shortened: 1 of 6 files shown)


hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
the above test was actually passed, It was just took some time for ceph validation to pass.

hkumarmk:
In fact this change is not affecting any code using for overcloud  - it is used in undercloud.

Issue #551, opened 2015-04-20
vgw test is disabled, as vgw creation is failing there in gcp


hkumarmk:
This need to be enabled once vgw issue is fixed

PR #685, opened 2015-06-24
fix vgw

hkumarmk:
VGW was not coming up because of missing ip_block configuration. This was
happend because of a hiera reference added for that param, which was changed
once, but missed to change the reference.


hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
This should fix issues #551 and 537

PR #686, opened 2015-06-25
Explicitely disable access to /run/shm/lttng* for vms

hkumarmk:
vm is trying to access /run/shm/lttng* file which is denied by apparmor, and
thus apparmor was throwing the denied logs in high volume. Explicitely denying
this should stop logging the incident.

Also this patch add apparmor package and service management code


hkumarmk:
please test this

hkumarmk:
@sorenh Could you please review this?

PR #688, opened 2015-06-26
test gate enviornment for ceilometer

punituee:
for bugs https://bugs.launchpad.net/jio/+bug/1453798
and https://bugs.launchpad.net/jio/+bug/1453797


punituee:
please test this

punituee:
do not merge

bodepd:
I have a general question about enabling ceilometer. Everyone's feedback that I've heard is that its performance is so terrible that it brings their entire cloud down and they ultimately have to turn it of. Do we have a way of understanding the potential performance impacts?

punituee:
I will be testing the same on the gate environment. It is currently our only option of generating bare minimum metrics. I will keep the events- pipeline and definitions to the minimal.
Some research on ceilometer yielded that it does a topic based subscription on the messaging queues and dumps the data in mysql. So an overloaded queue may be a problem.
We will never know unless we try.

punituee:
please test this

punituee:
please test this

PR #693, opened 2015-06-29
Create ssh key to be used by tempest to run ssh tests

pandeyop:
Need to create ssh key and provide the path in tempest.conf. This key would be used by tempest to run ssh tests.


manifests/tempest.pp
```diff
@@ -84,6 +84,20 @@
   }
 
   ##
+  # Generate ssh key to enable running ssh tests in tempest
+  ##  
+  ssh_key_tempest{'jenkins':
```
bodepd: this type needs to exist before you can use it. Perhaps start by using an exec resource with nova falvor-add command

manifests/tempest.pp
```diff
@@ -84,6 +84,20 @@
...
+    user => 'jenkins',
+    type => 'ssh-rsa',
+  }
+
+  exec { "generate_tempest_ssh_key":
+    command => 'ssh-keygen -b 4096 -t rsa -f /home/jenkins/.ssh/id_rsa -q -N ""',
+    creates => '/home/jenkins/.ssh/id_rsa',
+    require => user,
```
bodepd: require needs to reference a resource, you could add  Ssh_key_tempest['jenkins'], but then I think the order is backwards.

manifests/tempest.pp
```diff
@@ -84,6 +84,20 @@
...
 
   ##
+  # Generate ssh key to enable running ssh tests in tempest
+  ##  
+  ssh_key_tempest{'jenkins':
+    user => 'jenkins',
+    type => 'ssh-rsa',
+  }
+
+  exec { "generate_tempest_ssh_key":
```
bodepd: you might need to also create a file resource to update permissions to ensure that the tempest tests can access this user.

PR #696, opened 2015-07-02
UC | Make flavor management optional

hkumarmk:
Sometimes you may not need to manage nova flavors, so making it optional. This
is useful in case of undercloud controller where the flavors would have
extra_specs which are not supported by nova_flavor at this moment (we currently
use openstack client in nova_flavor which doesnt support setting extra_specs).


hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
@sorenh @bodepd please review this patch, this patch doesnt have any effect on overcloud, it only affect in undercloud controller.

PR #698, opened 2015-07-06
UC|Configure neutron only if ask to do so in tempest provision

hkumarmk:
In case of undercloud neutron is already present in the same machine and thus
reconfiguring neutron resources cause duplicate resource errors, and thus making
the configuration optional, with default to be on. Also adding an override in uc
role.


hkumarmk:
please test this

bodepd:
please test this

bodepd:
this build actually passed, it only failed b/c stmons took too long to stabalize.

PR #700, opened 2015-07-06
UC|Made lb optional

hkumarmk:
LB may not be always available, e.g undercloud does not have LB available. So
making it optional.


hkumarmk:
please test this

hkumarmk:
please test this

PR #702, opened 2015-07-07
Explicitely set some params to rjil::tempest

hkumarmk:
This patch is to set some params in rjil::tempest like passwords, which would likely to
be changed in different environments. Without this change, puppet run will be
failed in case of password changes.


hkumarmk:
please test this

PR #704, opened 2015-07-07
UC | Support overcast based deployments

hkumarmk:
This patch is to support overcast based deployments along with current one
(jiocloud.apply_resources). This patch introduce number of environment
variables by which deploy.sh will decide which way to deploy the system.

Here are the environment variables:

provisioner: if this is set to "overcast", overcast will be used for the
  provisioning. Default is current way of provisioning (jiocloud.apply resources)
overcast_yaml: this is the patch to overcast.yaml file where it configure
  different stacks and each steps in each stacks. Default one is the
  .overcast.yaml in current directory.
stack: Name of the stack to be deployed. Default is overcloud

overcast_source_repo: source repo for overcast
overcast_source_branch: source branch for overcast.

### DIFF
--- a/build_scripts/common.sh
+++ b/build_scripts/common.sh
@@ -28,14 +28,24 @@ if [ -n "${xtrace_was_on}" ]; then
   set -x
 fi
 
+##
 # Load map from generic image, flavor and network names to
 # cloud specific ids
+#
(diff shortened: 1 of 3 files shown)


hkumarmk:
please test this

PR #716, opened 2015-07-14
Disable TCP Offload on Builds

vpramo:
Patch to disable TCP offload on builds to ensure access to VM's launched on builds can be accessed via SSH

### DIFF
--- a/build_scripts/make_userdata.sh
+++ b/build_scripts/make_userdata.sh
@@ -107,6 +107,18 @@ echo 'cloud_provider='${cloud_provider} > /etc/facter/facts.d/cloud_provider.txt
 if [ -n "${slack_url}" ]; then
   echo 'slack_url=${slack_url}' > /etc/facter/facts.d/slack_url.txt
 fi
+
+##
+# Disable TCP Offloading in builds on Interfaces
+if [[ \$(facter is_virtual) == true ]];
+then
+  for nic in \$(ifconfig -a | awk '/eth[0-9]+/ {print \$1}'); do
+    ethtool -K \${nic} tx off
+    sed -i -e "/^exit 0$/i\ethtool -K \${nic} tx off" /etc/rc.local
+  done
+fi
+
+
 ##
 # Workaround to add the swap partition for baremetal systems, as even though
 # cloudinit is creating the swap partition, its not added to the fstab and not


vpramo:
please test this

vpramo:
@hkumarmk @sorenh  Please review and merge, the MTU size will be taken care of by the QA team using a cloud init script. I tested it and it works

manifests/system.pp
```diff
@@ -96,6 +96,15 @@
   }
 
   ##
+  # Disabling TCP offload on Virtual Environments only
+  if ::is_virtual {
+    exec { 'disable_tcp_offload':
+      command     => '/sbin/ethtool -K eth0 tx off',
```
hkumarmk: This should not be a part of puppet, this can be added in userdata (makeuserdata.sh).

I would run lshw (lshw -class network | grep logical.name) from userdata, and iterate through network interfaces, and just add ethtool -k ${nic} tx off also add the same in /etc/rc.local.

bodepd: couple other comments here if it were to be added to Puppet.

1. you need to tell Puppet when to run a command, it's not a system for launching commands, have a look at the creates, onlyif, unless parameters.
2. you can't hard-code eth0 here, you need to parameterize it so that the code itself doesn't make assumptions about things like what the name of the primary interface might be.

vpramo:
please test this

build_scripts/make_userdata.sh
```diff
@@ -107,6 +107,14 @@ echo 'cloud_provider='${cloud_provider} > /etc/facter/facts.d/cloud_provider.txt
 if [ -n "${slack_url}" ]; then
   echo 'slack_url=${slack_url}' > /etc/facter/facts.d/slack_url.txt
 fi
+
+##
+# Disable TCP Offloading in builds on Interfaces
```
bodepd: isn't this just supposed to be applied to VMs?

vpramo: Yes, only on build VM's, the idea is since we are doing regex for eth, it is valid only for Builds and not for HW nodes, as the interface types are em

hkumarmk: regex of eth is not going to make sure it is VMs. There can be baremetal servers, which have eth.

you could do the command "facter is_virtual" to verify f they are virtual or not.

vpramo:
please test this

jaspreetw:
HI @vpramo 

"the MTU size will be taken care of by the QA team using a cloud init script."
I don't agree to the above statement. Making this change to tempest test cases will be a big change, also we need  to maintain two repos for testing , one for production and one for gating env. 

I request you to find a alternate solution rather than changing the MTU size for each VM using cloud-init script

vpramo:
please test this

vpramo:
please test this

vpramo:
please test this

vpramo:
please test this

vpramo:
please test this

build_scripts/make_userdata.sh
```diff
@@ -107,6 +107,18 @@ echo 'cloud_provider='${cloud_provider} > /etc/facter/facts.d/cloud_provider.txt
...
+if [[ $(facter is_virtual) == true ]];
```
hkumarmk: Oh I see this is executing in jenkins - you would have to escape $  - \$(facter is_virtual)

build_scripts/make_userdata.sh
```diff
@@ -107,6 +107,18 @@ echo 'cloud_provider='${cloud_provider} > /etc/facter/facts.d/cloud_provider.txt
...
+then
+  for nic in `ifconfig -a | awk '/eth[0-9]+/ {print $1}'`; do
```
hkumarmk: Same here, you have to use $() rather than backtick

vpramo:
please test this

build_scripts/make_userdata.sh
```diff
@@ -107,6 +107,18 @@ echo 'cloud_provider='${cloud_provider} > /etc/facter/facts.d/cloud_provider.txt
...
   echo 'slack_url=${slack_url}' > /etc/facter/facts.d/slack_url.txt
 fi
+
+##
+# Disable TCP Offloading in builds on Interfaces
+if [[ \$(facter is_virtual) == true ]];
+then
+  for nic in \$(ifconfig -a | awk '/eth[0-9]+/ {print $1}'); do
+    ethtool -K ${nic} tx off
+    sed -i -e "/^exit 0$/i\ethtool -K ${nic} tx off" /etc/rc.local
```
bodepd: this keeps getting appended to the file, I found the following file from `bootstrap1-testjenkins-puppet-rjil-gate-1782`

````
ethtool -K  tx off
ethtool -K  tx off
ethtool -K  tx off
ethtool -K  tx off
ethtool -K  tx off
exit 0
````

File: build_scripts/make_userdata.sh [hunk 6]
```diff
@@ -107,6 +107,18 @@ echo 'cloud_provider='${cloud_provider} > /etc/facter/facts.d/cloud_provider.txt
 if [ -n "${slack_url}" ]; then
   echo 'slack_url=${slack_url}' > /etc/facter/facts.d/slack_url.txt
 fi
+
+##
+# Disable TCP Offloading in builds on Interfaces
+if [[ \$(facter is_virtual) == true ]];
+then
+  for nic in \$(ifconfig -a | awk '/eth[0-9]+/ {print $1}'); do
+    ethtool -K ${nic} tx off
```
bodepd: nic needs to be escaped here, it is being written into the userdata file as empty.

vpramo:
please test this

vpramo:
please test this

vpramo:
@hkumarmk @bodepd  The patch is verified on the build 1793, please merge.

jaspreetw:
please test this

PR #719, opened 2015-07-14
Fix sudo conf for user account which have dot in the name

hkumarmk:
sudo_conf break in case a user account have a dot in their username, and they
are not "admins". This is because sudo doesnt support dot or dash in the command
alias name. Fixing that problem here.


hkumarmk:
This issue is affected in production.

bodepd:
please test this

bodepd:
please test this

bodepd:
maybe we should put some test data in with this? I had to generate test data to validate that this works in the at env

PR #720, opened 2015-07-14
UC|more overcast specific changes

hkumarmk:
* Disabling debug as the code within debug function will not work with overcast.
* Fix wrong flavor reference
* unset env_http[s]_proxy before running make_userdata.sh
* Added validation stuffs for undercloud to .overcast.yaml
* it seems overcast has a bug because of which the variable is not getting
extracted properly on remote execution commands, because of which verify_hosts,
trigger_update etc are failing using overcast. Added an issue for that, for now
using python jiocloud directly from deploy.sh.
* save cleanup files in ~/ so they can be used by delete job.
* get_ip_of_node once rather than in a loop - ip will be assigned before that
get exeecuted

### DIFF
--- a/.overcast.yaml
+++ b/.overcast.yaml
@@ -36,7 +36,7 @@ overcloud:
       retry-wait: 5s
 undercloud:
   - shell:
-      cmd: "build_scripts/make_userdata.sh > userdata.txt"
+      cmd: "unset env_http_proxy env_https_proxy; build_scripts/make_userdata.sh > userdata.txt"
   - provision:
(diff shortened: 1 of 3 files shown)


hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

hkumarmk:
please test this

bodepd:
looks good as soon as it passes.

PR #728, opened 2015-07-16
UC | Params to disable tests that are not possible for undercloud

hkumarmk:
Provided params to rjil::test::neutron to diabled the tests to create networks
and fips as they are not possible to be run for undercloud.


hkumarmk:
please test this

PR #753, opened 2015-07-22
UC|Enable all interfaces in virtual environments

hkumarmk:
cloudinit only enable first interface even though there are multiple interfaces
configured in the machine. So in virtual environment, adding a userdata to
enable all interfaces with dhcp enabled, this is required for undercloud as
undercloud controller have multiple interfaces which need to be up.


hkumarmk:
please test this

hkumarmk:
please test me

PR #757, opened 2015-07-24
Modify tempest test list for gate environment

pandeyop:
This is the tempest tests lists that ought to run on the gate environment. The following tempest configuration values are also changed.
allow_tenant_isolation = false {Not running admin tests}
run_ssh = false {Nor running ssh tests till tempest is modified for MTU size fix}


pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

anshprat:
please test this

pandeyop:
please test this

anshprat:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

mayankkapoor:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

build_scripts/test.sh
```diff
@@ -2,8 +2,12 @@
...
 
+retval=0
 if [ ! -f ${project_tag}_ssh_config ]; then
   python -m jiocloud.apply_resources ssh_config --project_tag=${project_tag} ${mappings_arg} environment/${layout:-full}.yaml > ${project_tag}_ssh_config
+  retval=$?
 fi
 
 ssh -F ${project_tag}_ssh_config -l ${ssh_user:-jenkins} gcp1_${project_tag} '~jenkins/tempest/run_tempest.sh -N -- --load-list ~jenkins/tempest_tests.txt'
+
+exit $retval
```
anshprat: I thought the idea was to exit 0 (i.e to always pass) presently irrespective of the tempest test status?
Then why does the retval on line 8 matter?

build_scripts/test.sh
```diff
@@ -2,8 +2,12 @@
 
 . $(dirname $0)/common.sh
 
+retval=0
```
anshprat: This is not being used presently.

hiera/data/common.yaml
```diff
@@ -909,47 +909,51 @@ tempest::extra_config:
...
+  compute/env_type:
+    value: gate
+  compute/mtu_size:
+    value: 1400
   compute/use_floatingip_for_ssh:
     value: true
   compute-feature-enabled/disk_config:
     value: true
-  compute-feature-enabled/boot_from_volume_only:
-    value: true
```
anshprat: Why are these changes done? Would be good to have the reason for the changes in the commit msg.

hiera/data/common.yaml
```diff
@@ -909,47 +909,51 @@ tempest::extra_config:
...
+    value: False
   compute-feature-enabled/enable_instance_password:
-    value: true
+    value: True
   compute-feature-enabled/interface_attach:
-    value: true
+    value: True
   compute-feature-enabled/snapshot:
-    value: true
+    value: True
```
anshprat: Why has the case been changed? Would be good to maintain uniformity across files. Also, if this change has actually been done to maintain the uniformity, then would be good to have the reason for the changes in the commit msg.

build_scripts/test.sh
hkumarmk: I dont think that was the idea - we should have test.sh exit with return code according to the test status.

hkumarmk: But, here we are getting return value of ssh_config rather than tempest tests which is not right.

hiera/data/common.yaml
```diff
@@ -860,7 +860,7 @@ rjil::jiocloud::dhcp::dhcppools:
         mode: '0700'
 
     oncommit:
-      - set cip = binary-to-ascii(10, 8, ".", leased-address)
+      - set cip    value: binary-to-ascii(10, 8, ".", leased-address)
```
hkumarmk: This is not right, the original value is correct - this is to send appropriate value to the code,

hiera/data/common.yaml
```diff
@@ -830,8 +830,8 @@ dhcp::max_lease_time: 2592000
 #    gateway: 10.10.10.1
 #    oncommit_script: /tmp/script.sh
 #    oncommit:
-#      - 'set cip = binary-to-ascii(10, 8, ".", leased-address)'
-#      - 'set cmac = binary-to-ascii(16, 8, ":", substring(hardware, 1, 6))'
+#      - 'set cip    value: binary-to-ascii(10, 8, ".", leased-address)'
+#      - 'set cmac    value: binary-to-ascii(16, 8, ":", substring(hardware, 1, 6))'
```
hkumarmk: this is not correct, please revert to original one which is correct.

build_scripts/test.sh
pandeyop: @hkumarmk @anshprat: Anshu is right. We should have exit 0 irrespective of tempest test results i.e. failure of tempest tests should NOT fail the build. However, test.sh wouldn't *just* have tempest tests. This is why I tried to have *retval* return whatever was the result of tests *prior* to tempest.

hkumarmk: what is the purpose of always exit with 0 ? are we not gate with tempest test which was the motive when we start reducing the list to only relevant tests?

pandeyop:
please test this

pandeyop:
please test this

pandeyop:
please test this

anshprat:
@pandeyop (i) please squash the commits into a single commit (ii) please rebase on top of master.

pandeyop:
@anshprat This is done as per your suggestions.

PR #767, opened 2015-07-31
UC|Added a workaround to enroll all nodes

hkumarmk:
running enroll node code from dhcp is not working for some reason, which need to
be looked at, but until then this patch is adding a workaround to add a cron
which looks at the dhcp leases and enroll all nodes once in every day.


hkumarmk:
please test undercloud

bodepd:
needs to be rebased.

PR #768, opened 2015-08-01
Pausing puppet

anshprat:
Presently the only way to disable puppet is to disable the cron
This change provides for a way to check in config if puppet has to be run or not
Right now the python-jiocloud provider only has enabled_puppet True or False
Going forward, we will add the ability to "pause" puppet till a given time as well.
Though this remaining pausing change will only be python-jiocloud and nothing in puppet-rjil.

### DIFF
--- a/files/maybe-upgrade.sh
+++ b/files/maybe-upgrade.sh
@@ -3,29 +3,6 @@
 # When we're run from cron, we only have /usr/bin and /bin. That won't cut it.
 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
-##
-# Check if puppet is enabled for this host in consul
-# Exit codes:
-# 0 = Yes
-# 9 = No (Nein)
-##
-python -m jiocloud.orchestrate check_puppet
-puppet_enabled=$?
-
-if [ "$puppet_enabled" -eq 9 ]
-  then
-    echo "[WARN]: Puppet run is disabled, exitting."
-    exit 9
-fi
-
-# Exit codes:
-# 0: Yup, there's an update
-# 1: No, no updates
-# 2: Could not reach consul, so we don't know
-# 3: Could not reach consul, but we also haven't been initialised ourselves.
-python -m jiocloud.orchestrate pending_update
-rv=$?
-
 run_puppet() {
         # ensure that our service catalog hiera data is available
         # now run puppet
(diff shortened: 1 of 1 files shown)


anshprat:
This needs https://github.com/JioCloud/python-jiocloud/pull/64

anshprat:
please test this

anshprat:
anshup@bootstrap1-testjenkins-puppet-rjil-gate-1971:~$ python -m jiocloud.orchestrate enable_puppet false global 
false
anshup@bootstrap1-testjenkins-puppet-rjil-gate-1971:~$ python -m jiocloud.orchestrate check_puppet
anshup@bootstrap1-testjenkins-puppet-rjil-gate-1971:~$ echo $?
9

Aug  1 15:57:01 bootstrap1-testjenkins-puppet-rjil-gate-1971 CRON[1987]: (root) CMD (run-one /usr/local/bin/maybe-upgrade.sh 2>&1 | logger)
Aug  1 15:57:02 bootstrap1-testjenkins-puppet-rjil-gate-1971 logger: [WARN]: Puppet run is disabled, exitting.

anshprat:
I passed on 
export python_jiocloud_source_repo=git+https://github.com/anshprat/python-jiocloud
export python_jiocloud_source_branch=manage_config
to build the gate w/ the unmerged python-jiocloud changes.

files/maybe-upgrade.sh
```diff
@@ -3,6 +3,21 @@
...
+##
+# Check if puppet is enabled for this host in consul
+# Exit codes:
+# 0 = Yes
+# 9 = No (Nein)
+##
+python -m jiocloud.orchestrate check_puppet
+puppet_enabled=$?
+
+if [ "$puppet_enabled" -eq 9 ]
```
bodepd: I had assumed he would still run validation tests even if puppet is disabled.

anshprat: @bodepd so only run validation and not try to fix the broken validations right? This is doable. However, like you mentioned on the doc, only trying to fix validations without upgrade will need more work.

bodepd: I was saying still run the validation checks, but don't try to remediate, we can do the remediate only feature later.

bodepd:
please test this

files/maybe-upgrade.sh
```diff
@@ -37,6 +29,32 @@ validate_service() {
         fi
 }
 
+#Verify validation irrespective of enable_puppet
+validate_service
```
bodepd: puppet never runs the first time, validation runs and exits b/c it is in an initial failing state.

bodepd:
fixed as #778

PR #772, opened 2015-08-03
Dynamic tags for registering services for contrail upgrade procedure

vpramo:
For contrail upgrade, we will first do the upgrade contrail  controller on another bare-metal node and then cut over the compute nodes to the new contrail controller. When deploying the new node, we need to ensure that no traffic is sent to this node before the upgrade window. Hence a hiera override is done so that the service gets registered as backup and then during the upgrade the override will be changed to point to the new server.


hkumarmk:
This doesnt affect current systems. merging as it is urgent.

PR #773, opened 2015-08-03
[WIP] POC for upgrades

anshprat:


bodepd:
I'd love to chat more about this to understand what your motivations are here.

manifests/ceph/upgrade.pp
```diff
@@ -0,0 +1,72 @@
...
+  if ($consulkv['upgrade_component'] == 'ceph' and $consulkv['enable_upgrade'] == 'True' and $consulkv['enable_puppet'] == 'True'){
+    ##
+    # The below is not required for prod but is only for testing gates upgrade in gates
+    ##
+    if ($consulkv['gate_update_ceph_repo'] == 'True'){
+        file {'delete_ceph_giant_repo':
+          path   => '/etc/apt/sources.list.d/ceph.list',
+          ensure => absent,
+        }
+        apt_mirror::mirror { 'ceph_hammer_mirror':
```
bodepd: why is this being mirrored?

manifests/ceph/upgrade.pp
```diff
@@ -0,0 +1,72 @@
...
+          comment  => 'ceph hammer repo',
+          location => 'http://ceph.com/',
+          release  => 'debian-hammer',
+          repos    => 'main',
+          key      => {
+            'id'     => '17ED316D',
+            'server' => 'keyserver.ubuntu.com',
+          },
+        }
+        exec { "/usr/bin/apt-get -y upgrade":
```
bodepd: should we run this from puppet? maybe upgrade also does it.

manifests/ceph/upgrade.pp
anshprat: yeah, this is not required. was playing around w/ apt module initially. will remove.

manifests/ceph/upgrade.pp
```diff
@@ -0,0 +1,72 @@
...
+          refreshonly => true,
+          subscribe => File["/etc/apt/sources.list.d/ceph_hammer_repo.list"],
+          timeout => 3600,
+        }
+    }
+
+        if (('stmonleader1' in $::hostname) and ($consulkv['mon_restart_done'] != 'True')){
```
bodepd: I don't get how this works, where is the key being set?

manifests/ceph/upgrade.pp
anshprat: well, this is to control / keep upgrade within the same flow. As per this , we upgrade in the same cycle. Else for maybeupgrade, we need to go out of this and do upgrade in the next cycle.

manifests/ceph/upgrade.pp
anshprat: The initial keys will be set similar to trigger_update. For the rest of the nodes, the nodes that complete their own upgrade, will update the key for the next node.

PR #777, opened 2015-08-04
Reducing noise - userdata

anshprat:
Removing 'v' option from cp and tar in make_userdata
These are not generally helpful in the present state.
Ideally there could be a debug/verbose flag being passed at build creation
which should enable such verbosity. Would add such a flag later.


bodepd:
haha. finally :)

bodepd:
please test this

PR #781, opened 2015-08-10
Support for Configurable contrail-dns for v2.1

vpramo:
When upgrading to v2.1 contrail, contrail deploys bind, this conflicts with the existing dnsmasq which we use for consul. Hence to avoid this, we are going to use two approaches, 1. disable contrail-dns by default as we are not using the service now, 2. we are going to add a config to the existing dnsmasq to forward dns traffic based on VM_domains to contrail-dns when enabled.


hiera/data/common.yaml
```diff
@@ -670,9 +670,15 @@ contrail::vrouter::manage_repo: false
...
 contrail::enable_analytics: false
 
+contrail::enable_dns: false
+contrail::dns_port: 10000
+
 contrail::vrouter::debug: true
 
 rjil::contrail::server::enable_analytics: false
+rjil::contrail::server::enable_dns: "%{hiera('contrail::enable_dns')}"
+rjil::contrail::server::vm_domain: undef
```
bodepd: you don't need to reset things in hiera that are the same as the defaults that you are supplying. also, I don't think that undef is treated as undef here, you need to use the symbol (I think it gets treated as a string like this)

manifests/contrail/server.pp
```diff
@@ -19,6 +22,14 @@
   }
 
   include ::contrail
+  if $enable_dns and $vm_domain {
+    include dnsmasq
+    dnsmasq::conf { 'contrail':
+      ensure  => present,
+      content => "server=/${vm_domain}/127.0.0.1#${dns_port}",
```
bodepd: doesn't bind scale way better than dnsmasq? This solution seems a little naive for once we actually want to use contrail for DNSAAS

vpramo: Eitherway works, we can have a forwarder in contrail-named to forward to consul as well. Right now we are using dnsmasq as a forwarder between consul DNS and contrail-Dns service. So essentially in terms of scale not sure where the issue comes.

bodepd: won't the traffic for DNSAAS requests come in on port 53 which will go through dnsmasq?

vpramo: Yes but its pretty much doing forwarding and nothing else...we can try some heavy stress testing to see if it scales.. only on forwards.. if not we need to reverse this i.e forward via contrail-named.

bodepd:
@sorenh any comments about this? Do you think it's worth worrying about?

hiera/data/common.yaml
vpramo: @bodepd sure.. i will remove, i just wanted to keep it for reference thats all.

sorenh:
What happens if we just don't deploy contrail-dns?

bodepd:
@sorenh that is the behavior that is being proposed, but there is additional code that keeps dnsmasq on port 53 and has it forward to bind on another port

PR #782, opened 2015-08-14
Enable contrail analytics on a different node

vpramo:
Patch to have a separate node  for Contrail analytics and disabling contrail analytics on current contrail controller

### DIFF
--- a/manifests/contrail/analytics.pp
+++ b/manifests/contrail/analytics.pp
@@ -1,0 +1,31 @@
+###
+## Class: rjil::contrail
+###
+class rjil::contrail::analytics 
+ {
+  ##Include contrail collector
+  class {'::contrail':
+    enable_analytics => true,
+    enable_config    => false,
+    enable_control   => false,
+    enable_webui     => false,
+    enable_ifmap     => false,
+    config_ip        => join(service_discover_dns('real.neutron.service.consul','name')),
+  }
+
+  # Added Tests
+  rjil::test {'contrail-analytics.sh':}
+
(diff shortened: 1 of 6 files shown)


manifests/contrail/analytics.pp
```diff
@@ -0,0 +1,32 @@
...
+class rjil::contrail::analytics 
+ {
+  ##Include contrail collector
+  class {'::contrail':
+    enable_analytics => true,
+    enable_config    => false,
+    enable_control   => false,
+    enable_webui     => false,
+    enable_ifmap     => false,
+    config_ip        => join(service_discover_dns('real.neutron.service.consul','name')),
```
sajuptpm: What is the difference b/w neutron running in ct1(real) and haproxy1 ?

site.pp
```diff
@@ -80,6 +80,15 @@
   include rjil::neutron::contrail
 }
 
+##
+## Setup Contrail Analytics Nodes
+node /^ca\d+/ {
+  include rjil::base
+  include rjil::redis
+  include rjil::cassandra
```
bodepd: does this need it's own cassandra cluster?

manifests/contrail/analytics.pp
sajuptpm: go it, real and lb service. I didn't check the process.

site.pp
vpramo: I expect analytics not to interfere with contrail control DB.. hence this... but we can still have the data on the contrail control cassandra as well.

MayankGo:
Pramod who has asked you to do this activity......

PR #787, opened 2015-08-24
stdlib upgrade + puppet-consul revision

anshprat:
Upgrading stdlib as one of the new commits in puppet-consul
needs stdlib nees validate_integer which is in stdlib >4.6.0
Also, fixing the puppet-consul version to latest revision
to avoid such surprises going further


anshprat:
please test this

PR #795, opened 2015-09-07
Better Support autodetecting ceph osd disks

hkumarmk:
Old version of autodetect code used to detect all disks including system disks
which was supposed to be used for operating system. This patch added:

* custom facts system_disks with list of disks which are connected to builtin
controllers and nonsystem_disks fact which have list of disks which are connected
to jbod.
* Custom fact nonsystem_blankorcephdisks which have Nonsystem disks which are either
ceph or blank disks. It use /dev/disk/by-path/ to detect non-system disks. Usually
baremetal system would have some system disks which are connected to builtin controllers
and disks connected to JBODs which are connected through pci disk controllers.
* Removed parameter autodetect from rjil::ceph::osd, as it is not required,
rather it is implemented by using either custom fact (autodetect) or a list of
disks.


hkumarmk:
please test this

anantkaushik89:
please test this

anantkaushik89:
please test this

PR #806, opened 2015-09-29
WIP | Containers

hkumarmk:
Note: Some temporary changes have been done just to make this patch working which would be changed once this is ready


Vagrantfile
```diff
@@ -71,11 +71,21 @@ Vagrant.configure("2") do |config|
       config.vm.provision 'shell', :inline =>
       "echo env=#{environment} > /etc/facter/facts.d/env.txt"
 
-      if ENV['http_proxy']
```
bodepd: can we just merge these vagrant changes?

hkumarmk: I can raise a separate patch if we can merge them

docker_build/containers/base/maintain.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+#
+exec 2>&1
+
+wait_time=${wait_time:-10}
+run_puppet=1
+
+##
+# if consul_discovery_token is set, write it to factor
```
bodepd: facter is mispelled

docker_build/containers/base/maintain.sh
```diff
@@ -0,0 +1,55 @@
...
+# fact for env
+if [ -n $env ]; then
+  echo 'env='${env} > /etc/facter/facts.d/env.txt
+fi
+
+##
+# Below chunk of code will run in infinite loop which make sure this container
+# and apps running in this container is up and running and configured correctly.
+##
+while true; do
```
bodepd: why does this run in a loop? What keeps track of when the containers are ready?

docker_build/containers/base/site.pp
```diff
@@ -0,0 +1,4 @@
+Exec { path => [ "/bin/", "/sbin/" , "/usr/bin/", "/usr/sbin/", "/usr/local/bin/","/usr/local/sbin/" ], logoutput => true }
```
bodepd: do you have to split these out into their own files? is there an advantage? Why not just leave them in a single site.pp?

docker_build/scripts/build_all.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+dir=$(dirname $0)
+cd $dir
+# build base image first
+./docker_build.sh
+for i in `ls ../containers` ; do
```
bodepd: does this create a container called system?

docker_build/scripts/build_all.sh
```diff
@@ -0,0 +1,32 @@
...
+    keystone)
+      ./docker_build.sh keystone 5000,35357 10
+      ;;
+    db)
+      ./docker_build.sh db 3306 10
+      ;;
+    memcached)
+      ./docker_build.sh memcached 11211 10
+      ;;
+    consul)
```
bodepd: why do you have a manifest for docker and system is they are not being built? This patch is already too big, anything that we can do to eliminate this complexity will help with code review.

docker_build/scripts/entrypoint.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
bodepd: what is the difference between this and maintain.sh?

environment/docker.yaml
```diff
@@ -0,0 +1,100 @@
...
+    networks: default
+    boot_volume: default_volume
+  bootstrap:
+    number: 1
+    flavor: small
+    image: trusty
+    networks: default
+    boot_volume: default_volume
+    assign_floating_ip: true
+  kdb:
```
bodepd: why are you using an entire physical host for the db?

environment/docker.yaml
```diff
@@ -0,0 +1,100 @@
...
+    image: trusty
+    networks: default
+    boot_volume: default_volume
+  keystone:
+    number: 1
+    flavor: small
+    image: trusty
+    networks: default
+    boot_volume: default_volume
+  gdb:
```
bodepd: why do you have different dbs for keystone and glance?

hiera/data/common.yaml
```diff
@@ -110,23 +116,24 @@ public_network_name: public
...
-rjil::openstack_objects::users:
+keystone_db_host: kdb.jiocloud.com
+keystone_memcached_host: kcache.jiocloud.com
+rjil::keystone::objects::users:
   "%{hiera('test_user_name')}":
     password: "%{hiera('test_user_password')}"
     user_tenant: "%{hiera('test_user_tenant')}"
     create_network: false
 ##
-# rjil::openstack_objects::tenants can be either a array of tenant names or a
```
bodepd: I agree that object creation needs to be split out, it should be done separately from this patch. This work is more or less the intent of this patch. Maybe we can consolidate these efforts since they are trying to solve the same thing.

hiera/data/common.yaml
```diff
@@ -461,11 +470,11 @@ rjil::keystone::ssl: true
 rjil::keystone::protocol: "%{hiera('api_protocol')}"
 rjil::keystone::version: v2.0
 rjil::keystone::admin_email: "%{hiera('admin_email')}"
-rjil::keystone::public_port: "%{hiera('keystone_port')}"
-rjil::keystone::public_port_internal: "%{hiera('keystone_localbind_port')}"
+rjil::keystone::userapi_port: "%{hiera('keystone_port')}"
```
bodepd: why are these variables being changed, is this somehow related to the docker changes?

hiera/data/common.yaml
```diff
@@ -1022,3 +1067,12 @@ squid3::refresh_patterns:
...
 rjil::haproxy::default_timeout_server: 50000
+
+#### docker
+docker::package_name: docker-engine
+docker::manage_package: false
+docker::manage_kernel: false
+docker::use_upstream_package_source: false
+docker::extra_parameters: ["--insecure-registry=192.168.100.4:5000"]
+docker_image_version: "%{image_version}"
+rjil::docker::registry_url: 192.168.100.4:5000
```
bodepd: is this vagrant specific?

hiera/data/hw_type/docker.yaml
```diff
@@ -0,0 +1,11 @@
+# Service supervisor is required in docker containers for services which use upstart, as upstart doesnt work in docker containers.
+rjil::keystone::service_manager: runit
```
bodepd: what is the advantage of even using a service manager in a container as opposed to just managing the exec?

manifests/db.pp
```diff
@@ -13,6 +13,8 @@
   $mysql_data_disk = undef,
   $dbs = {},
   $bind_address = '0.0.0.0',
+  $monitor_user = 'monitor',
```
bodepd: what does a monitor user have to do with container support?

manifests/docker/container.pp
```diff
@@ -0,0 +1,145 @@
...
+  $mount_service_check_volume  = true,
+  $consul_check_scripts   = undef,
+  $consul_check_ttls      = undef,
+  $consul_check_intervals = '10s',
+  $consul_service_names   = $name,
+  $consul_service_tags    = undef,
+) {
+
+  ##
+  # Mount /usr/lib/jiocloud/tests/service_checks across the containers so consul will be able to run
```
bodepd: why not just install the service checks on the host, all that consul really need to know to run service checks is what port/interface the containers are going to be launched on? It looks like running the service checks on the containers themselves adds a ton of complexity?

manifests/docker/container.pp
```diff
@@ -0,0 +1,145 @@
...
+      ensure => 'directory',
+    }
+
+    $log_volume = ["/var/log/containers/${name}:/var/log/"]
+  } else {
+    $log_volume = []
+  }
+
+  ##
+  # merge common volumes and volumes - common volumes is the volumes which will
```
bodepd: is this really a workflow that we are going to encourage? I'd like to get more data here on what the use cases are.

manifests/docker/container.pp
```diff
@@ -0,0 +1,145 @@
...
+  # environment to share the development folder between docker host and
+  # container so changing the data in docker host will cause the change added on
+  # all containers in the docker host. Combining with vagrant it is going to be
+  # very helpful.
+  ##
+  $all_common_volumes = concat($common_volumes, $log_volume, $jiocloud_volume)
+  $all_volumes = union($all_common_volumes, $volumes)
+
+  ##
+  # if consul_service_names is a hash, then create multiple environment
```
bodepd: looking at this code, I am having trouble understanding what exactly it does. I think you are passing host configs to the containers so that the code to monitor them can be used on the containers? Doesn't consul just wind up doing the monitoring? Why not just run consul on the host, and just have code that adds the checks to the base host?

manifests/glance.pp
```diff
@@ -1,4 +1,17 @@
 ## Class: rjil::openstack::glance
+#
+# [*service_manager*]
+#   Service manager, by default it is using system specific service manager, say
+#   for ubuntu it is upstart. But in case of docker container, docker doesnt
+#   support upstart service scripts, so have to use different one, another
+#   supported service manager is 'runit'.
+# [*service_registrator*]
```
bodepd: why is this a good thing? It seems to add a ton of complexity and no value, or am I missing something?

manifests/glance.pp
```diff
@@ -130,14 +190,23 @@
...
-  }
 
-  rjil::jiocloud::consul::service { 'glance-registry':
-    tags          => ['real'],
-    port          => $::glance::registry::bind_port,
+  ##
+  # In case of running glance in container, there is one app called registrator
+  # which can be used to automatically register/deregister the consul services
+  # automatically as and when the container up/down. in which case no need to
+  # use rjil::jiocloud::consul::service to register the service.
```
bodepd: again, the old method of just adding a config file to consul running on the host seems way easier than adding an entirely different container just to bridge between consul and docker, or am I missing something?

manifests/keystone.pp
```diff
@@ -19,40 +33,81 @@
...
+      port          => 35357,
+    }
   }
 
-  rjil::jiocloud::consul::service { "keystone-admin":
-    tags          => ['real'],
-    port          => 35357,
+  ##
+  # runit can be used to supervise services. This is useful especially for
+  # containers when you run multiple processes.
```
bodepd: are we going to run multiple processes from the same container? Doesn't that violate the design goal of microservices?

manifests/keystone.pp
```diff
@@ -19,40 +33,81 @@
...
+  ##
+  if $service_manager == 'runit' {
+    rjil::runit::service {'keystone':
```
bodepd: I'd like to understand why runit is better than just directly having docker invoke keystone-all?

manifests/keystone/objects.pp
```diff
@@ -0,0 +1,56 @@
+#
```
bodepd: this should be a different pull request, we should have a separate requirement of decoupling deps from services (ie: breaking down oc role) and mark it as a dep for this project so that it can be done in the next sprint.

manifests/openstack_objects.pp
```diff
@@ -8,9 +8,6 @@
 class rjil::openstack_objects(
   $identity_address,
   $override_ips      = false,
-  $users             = {},
```
bodepd: should we just eliminate this class and have it's config moved to the relevent keystone/neutron/glance object creation classes?

manifests/system_container.pp
```diff
@@ -0,0 +1,37 @@
...
+  # of a misconfiguration of subnets, those facts can return null values for
+  # those facts. In that case, puppet run should explicitely fail to avoid the
+  # system to go into invalid state.
+  # NOTE: This code assume the fact is only used for private_address,
+  # public_address, private_interface, public_interface. If in future those
+  # facts used for anything else, they must be validated either here or in
+  # appropriate code.
+  ##
+
+  if $check_address_facts {
```
bodepd: I may need to discuss this with you, I couldn't exactly understand what it does.

bodepd:
a few high level notes, some of this is repeated form the inline comments that I made, but I wanted to try to summarize. This patch is actually related to three types of changes:

1. docker related changes
2. changes to vagrant
3. changes to decouple config of services (pretty much the same as: https://github.com/JioCloud/puppet-rjil/pull/793/files)

This patch seems way too complicated, and I'd like to better understand the motivations for some of the changes:
1. moving from site.pp to per service manifests (lots of code changes, not sure why we can't use site.pp)
2. using some service to proxy between consul and services, adds a ton of complexity, and I don't understand what the advantage is. From a consul perspective, I'd say that port/service checks are how it should integrate with services, and it looks like running consul on the host, and using puppet to make it away of what containers to monitor (via their ports) is a pretty simple solution that achieves what we want. Your proposed solution adds at least a hundred lines of code, splits us into two consul models (host vs. whatever this other one is) and seems pretty complicated ( I could not fully grok it)
3. not sure what the advantage of runit is (although it seems likely that is will be a better solution that running services from docker)

hiera/data/common.yaml
hkumarmk: even though the IP seems to be vagrant IP, but it is going to be there for longer. We are going to push built containers to a registry and we would need the registry url to be specified, eventually it is going to be a url from where everybody get the containers.

hkumarmk:
I am putting the answers below, but it need more detailed discussion. The intention as of now is to reuse the deployment code as much as possible, and later we can use different other things to reduce the puppet complexity - rather than puppet do everything, we could use other projects where it make sense - like we may try using kubernets/messos for container placements, use micro operating systems like coreos/ranchoros/snappy etc

docker related changes
changes to vagrant

Harish: This change can be separated, (its just a single commit, the reason I did not sepaate it because now a days we are not merging any code, and without merging that code I could not continue)

changes to decouple config of services (pretty much the same as: https://github.com/JioCloud/puppet-rjil/pull/793/files)

Harish: In order to get this code work, we need to split the object creation and other stuffs as they are going to be running on different nodes, so I had to do it. I thought to reduce the complexity and reduce the common code - like rather than putting everything in rjil::openstack_objects, I thought to split them under specific prorject specific code like rjil::keystone::objects. Again now a days we are not merging any code, so I did not split it as separate patch atm.

This patch seems way too complicated, and I'd like to better understand the motivations for some of the changes:
1. moving from site.pp to per service manifests (lots of code changes, not sure why we can't use site.pp)
 
Harish: Here are the reasons I thought of

a. Once move to container (with net=host), every container name will be same as host's name, so you cannot have node level blocks, so you may have to have code blocks with container name (its another factor I push to containers) conditionals.
b. I thought these individual pieces would be handled by individual team, so it would be simpler for them to have very simple flat site.pp, otherwise number of containers are going to be large in number, and you are going to see lot of conditionals there.

2. using some service to proxy between consul and services, adds a ton of complexity, and I don't understand what the advantage is. From a consul perspective, I'd say that port/service checks are how it should integrate with services, and it looks like running consul on the host, and using puppet to make it away of what containers to monitor (via their ports) is a pretty simple solution that achieves what we want. Your proposed solution adds at least a hundred lines of code, splits us into two consul models (host vs. whatever this other one is) and seems pretty complicated ( I could not fully grok it)

Harish: I actually just tried this option initially to see how it work. Here is the major difference i see once we move to containers.
a. The host will not have anything but docker installed and a code to provision appropriate containers and  NOTHING ELSE.
b. everything else would be running on its own containers (may be with multiple services in same containers or individual containers in the services - my initiative in that is the last commit which is again just work in progress)
c. So now I think the complexity increase in terms of maintanance if we use putting service configs in the container where the consul runs - what we do is that we put service configs on the consul container and will reload consul there. I initially thought to have those service scripts installed in the consul container itself, then I have seen this "registrator" which monitor docker socket and register/deregister the services on consul automatically. 
d. Even though it is introducing one container on every host, I feel it is adding values as below
  i. Now all you need is to start container and nothing else (no need to place service specific configs in consul container, and no need to restart consul service in consul containers) - your service is started in the container and got registered in consul with no extra steps.
  ii. Other real advantage I see when we do a upgrades or doing some maintainance on the containers and we need to test it before putting to live pool. Now let me consider a scenario on we upgrade say nova-api. We may start the container (initially) with different name (say test.nova.service.consul, rather than real.nova.service.consul) so we can test it before moving to real.nova.service.consul. 

   With puppet based solution, what we need to do is that, start nova-api container, put service config file with tag 'test' on consul container, reload consul on consul container (this need to be coordinated as they are running in two different system) and then test.
   Once test is okay, then reconfigure the service config file with tag 'real', and reload the consul on consul container (again need to be coordinated).

   Note that, we can use shared mounts to place service configs on consul containers, but still service reload of consul need to be happen on the consul container itself.

    With registrator based solution, what we do is to start container with tag 'test' initiallly, test it, once we are okay, restart the container with tag 'real' - no need of coordination, no going back to other (consul) container if we need to make any changes/create container

   iii. As mentioned on ii, reduced complexity in certain situations by just a self contained/self managed containers, no need to go back to consul container to do the service management, once we have shared mount for service scripts placed (or if the service scripts are available in consul container) we are done - just start/stop containers, your services will get registered/deregistered automatically.


3. not sure what the advantage of runit is (although it seems likely that is will be a better solution that running services from docker)

Harish: Initially I tried using direct command run as entry point (like keystone-all), but then I see certain problems as below

a. we run validation checks to validate service status which is actually beyond just seeing the process is running - where what docker make sure if we run direct service command is that the process just run. But we are going to have certain situations where service go bad even though process is running. So we need validation running every time, and if failed validation then to 1. report it back to consul (this patch is not doing that atm, which will be done before it ready for serious review), 2. fix the issues by running puppet. So I believe validation checks is a must thing to do

b. Now next option is to run validation on base host or a dedicated container, I thought that is going to add more complexity as 
  i. either we would be running puppet on that dedicated container or base host to configure container specific configuration, and run validation, which is going to fail the purpose of this effort of trying to separate the stuffs out - say for example, if nova-compute is going to upgrade which need certain changes in puppet where the container which run contrail-vrouter-agent is not upgraded. So I would suggest to have each container self sufficient to run everything it need to be running from that container itself.

c. Then I tried created a script to run validation and puppet runs in background (in loop as cron is not going to be running inside the container normally) and run keystone-all in foreground so docker will handle keystone. But then I see same problem as mentioned in (a), running validation in background cannot handle the process running in foreground (the moment it is down, the whole container is going to be end.

d. Then I tried in reverse, run main service (like keystone) in background and validation script in frontend. But then I face a problem of upstart not being supported in docker - you cannot run upstart scripts in docker container (we have all openstack services using upstart scripts). 

e. And I see the advantages of runit - 1. it is the process supervisor which just restart process if its down which is better than verifying the process using the process id get from a pid file (like upstart and other service scripts). 2. at the same time it provide a way to manage the service like service scripts for individual services (and is integrated with puppet). Other thing is that runit can do process logging which is configuable - all you need is the process to send the result to stdout/err, runit will log them in individual files - like upstart, but in upstart it is not configuable, say for example, same logs will be printed on upstart logs as welll as keystone logs, in case of runit, what I do is to not configured logging from runit in case of keystone and glance but is enabled for maintainance.sh script (which run validation and if required puppet run)

f. Other thing I am trying to solve is that, trying to avoid any zombie processes - which may happen if the service command we run doesnt handle reaping its subprocesses in case of a terminiation (during docker stop). This would need bit more work, somebody has done this work already, (I would be taking certain stuffs from them to fix this problem completely). (https://github.com/phusion/baseimage-docker)

PR #807, opened 2015-09-29
Fix for tftpboot directory permission issues

hkumarmk:
the user which run tftp should have enough permissions for /tftpboot and its
subdirectories (read only permission is required). So fixing it.


manifests/ironic.pp
```diff
@@ -37,21 +37,26 @@
...
     ensure => 'directory',
     owner  => 'ironic',
     group  => 'ironic',
+    mode   => '0755',
   } ->
   file { '/tftpboot/pxelinux.0':
+    ensure  => 'file',
     owner   => 'ironic',
     group   => 'ironic',
+    mode    => '0644'
```
anshprat: missing comma

anshprat: ---> syntax:manifests

Could not parse for environment production: Syntax error at 'source'; expected '}' at /home/travis/build/JioCloud/puppet-rjil/manifests/ironic.pp:53

rake aborted!

PR #809, opened 2015-09-30
Changed Vagrantfile to work behind proxy as of deploy.sh

hkumarmk:
* Use env_http_proxy/env_https_proxy as we use in cloud based provisioning
* Added proxies for both of them to work for all connections
* Changed puppet config with right default manifest and other basic configs

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -71,10 +71,20 @@ Vagrant.configure("2") do |config|
       config.vm.provision 'shell', :inline =>
       "echo env=#{environment} > /etc/facter/facts.d/env.txt"
 
-      if ENV['http_proxy']
+      if ENV['env_http_proxy']
         config.vm.provision 'shell', :inline =>
-        "echo \"Acquire::http { Proxy \\\"#{ENV['http_proxy']}\\\" }\" > /etc/apt/apt.conf.d/03proxy"
+        "echo \"Acquire::http { Proxy \\\"#{ENV['env_http_proxy']}\\\" }\" > /etc/apt/apt.conf.d/03proxy"
+        config.vm.provision 'shell', :inline =>
+        "echo http_proxy=#{ENV['env_http_proxy']} >> /etc/environment"
       end
+
+      if ENV['env_https_proxy']
+        config.vm.provision 'shell', :inline =>
+        "echo https_proxy=#{ENV['env_https_proxy']} >> /etc/environment"
+      end
+
+      config.vm.provision 'shell', :inline =>
+        "echo no_proxy='127.0.0.1,169.254.169.254,localhost,consul,jiocloud.com' >> /etc/environment"
 
       # run apt-get update and install pip
       unless ENV['NO_APT_GET_UPDATE'] == 'true'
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -71,11 +71,21 @@ Vagrant.configure("2") do |config|
       config.vm.provision 'shell', :inline =>
       "echo env=#{environment} > /etc/facter/facts.d/env.txt"
 
-      if ENV['http_proxy']
+      if ENV['env_http_proxy']
```
anshprat: @hkumarmk why a separate env_http_proxy instead of using existing http_proxy. That way folks dont have to export it separately. The existing http_proxy on the bm can be picked and used directly.

anshprat: the one time I think this would be useful is when host and vagrant have difference in this setting, but in our case, that wont be the case on bm.

hkumarmk: no what I used to do is to configure proxy on vagrant host, with upstream proxy config. And use that proxy for vms. 

Advantage: provided you have appropriate cache config in proxy, lot of your packages etc would be served locally which will improved the deployment performance.

anshprat: hmm, yeah, another usecase. But not of much help when we are running everything within the DC. (local vagrant proxy cache vs prod dc cache)

hkumarmk:
Now this should handle both the situations.

anshprat:
looks good now. will check it locally and merge.

Vagrantfile
```diff
@@ -71,11 +71,24 @@ Vagrant.configure("2") do |config|
       config.vm.provision 'shell', :inline =>
       "echo env=#{environment} > /etc/facter/facts.d/env.txt"
 
-      if ENV['http_proxy']
+      http_proxy = defined?(ENV['env_http_proxy']) ? ENV['env_http_proxy'] : defined?(ENV['http_proxy']) ? ENV['http_proxy'] : nil
```
anshprat: it seems to work with env_http_proxy but not http_proxy
anshup@vagrant01:~/puppet-rjil$ ruby
a=defined?(ENV['env_http_proxy']) ? ENV['env_http_proxy'] : defined?(ENV['http_proxy']) ? ENV['http_proxy'] : nil
puts a

anshup@vagrant01:~/puppet-rjil$ echo $http_proxy
http://10.140.192.34:3128/

hkumarmk:
hmmm it seems ruby assume that environment variables are defined (even though its not defined in the shell). I changed it to check if the variable is empty or not, which should work.

(http_proxy = !ENV['env_http_proxy'].to_s.empty? ? ENV['env_http_proxy'] : !ENV['http_proxy'].to_s.empty? ? ENV['http_proxy'] : nil)

anshprat:
Can we do just  http_proxy = ENV['env_http_proxy'] || ENV['http_proxy'] ? any pitfalls of this? This looks much simpler and easier compared to the empty check.

hkumarmk:
This should work :) I just did not check that

hkumarmk:
@anshprat pushed as you suggested

PR #813, opened 2015-10-02
Adding SSH key

gdpak:
Add SSH keys


anshprat:
Hi Deepak,
The last revert was on the entire PR. So you will need to make changes to add the user account in at.yaml as well again.

anshprat:
Also, after adding the at.yaml, please add it to the same commit (git commit --amend) to keep a single commit msg.

anshprat:
looks good. +1 for merging. I would let a second pair of eyes confirm and merge.

mayankkapoor:
Looks fine to me. Merging.

PR #838, opened 2015-11-06
Kilo cinder backup limit

swamireddy:


anshprat:
Need to check why is jiojenkins commenting here :!
Looks good. Merging this now as its in kilo branch and will not impact existing stuff. Before merging kilo to master, we would need to get it gated and AT.

swamireddy:
I could see the jiojenkins commenting for my previous patches also... I guess, some one might have added for comment for all patches

PR #841, opened 2015-11-06
Added Puppet-omd Module with its Dependency

bhavinajmera:
Have added Puppet-Omd module.


anshprat:
You need to add only the changed files. You have added all the modules.

PR #845, opened 2015-11-13
Fixed error in Vagrantfile

vpramo:
We need to copy the hiera.yaml to /etc/puppet else hiera will not be referenced


anshprat:
Vagrant only change, this was there earlier. so adding it back now.

PR #851, opened 2015-11-24
Adding Postfix Module

bhavinajmera:
As discussed I have created a postfix module for SMTP relay.

### DIFF
--- a/manifests/commonservices/postfix.pp
+++ b/manifests/commonservices/postfix.pp
@@ -1,0 +1,40 @@
+class postfix (
+$myhostname,
+$mydomain,
+$mydestination,
+$relayhost,
+) {
+
+package {'postfix':
+  ensure => 'installed',
+  }
+
+file {'/tmp/bhavin':
+  content => "
+smtpd_banner = $myhostname ESMTP $mail_name (Ubuntu)
+biff = no
+append_dot_mydomain = no
+readme_directory = no
+smtpd_tls_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem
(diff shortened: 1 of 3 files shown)


manifests/commonservices/postfix.pp
```diff
@@ -0,0 +1,40 @@
...
+myorigin = /etc/mailname
+mydestination = $mydestination
+relayhost = $relayhost
+mynetworks = 127.0.0.0/8 [::ffff:127.0.0.0]/104 [::1]/128
+mailbox_size_limit = 0
+recipient_delimiter = +
+inet_interfaces = loopback-only
+inet_protocols = all
+",
+  require => Package['postfix']
```
vpramo: This will just create a file in /tmp/bhavin , how will postfix pick this config file, secondly better to move this to a template.

manifests/commonservices/postfix.pp
```diff
@@ -0,0 +1,40 @@
+class postfix (
+$myhostname,
```
vpramo: better to have defaults

site.pp
```diff
@@ -223,4 +223,5 @@
   include rjil::commonservices::jenkins::master
   include rjil::commonservices::tools
   include ::omd::server
+  include postfix
```
vpramo: you need to give the complete path of your class, so it will be rjil::commonservices::postfix

manifests/commonservices/postfix.pp
```diff
@@ -0,0 +1,40 @@
...
+$mydomain,
+$mydestination,
+$relayhost,
+) {
+
+package {'postfix':
+  ensure => 'installed',
+  }
```
vpramo: i thought we were using a module for postfix right

PR #852, opened 2015-11-30
Fixing some of the common services requirements:

anshprat:
i) Added apache user as it may not be present based upon who/what/how apache was installed.
ii) Added hiera for public/private interface as presently its requirement is hardcoded.
iii) Changing the default location of jenkins gitconfig to default jenkins home.
iv) Made consul as server role in tools boxes.
TODO: Need to add bootstrap role - presently it has been added manually to tools01 hiera.


mayankkapoor:
@akash1808 reviewed it. Merging.

PR #861, opened 2016-01-07
Install haveged on tools

mayankkapoor:
This will allow RSA keys generation much faster


mayankkapoor:
@anshprat Please review.

anshprat:
@mayankkapoor the PR is good but it has merges within the commits. Based upon experience, doing a revert of such merges (which contain other merge) is a royal pain. So would recommend doing a clean PR with only a single commit related only to this PR. Also, we should test this on vagrant once to ensure there is no package conflicts.

mayankkapoor:
@anshprat Squashed the commits into a single commit.

